PR #1, opened 2013-01-21
Capitalize OR

wraithan:
Proper and easier to read because it isn't mistakable for the word `or`.


ericholscher:
:horse: 
:fire:

PR #3, opened 2013-01-25
Introduced 'about' folder to store content related to the project itself

benoitbryon:
Proposition is to store all information about the project itself in a "about" folder.
I also introduced some standard files such as LICENSE, AUTHORS and CHANGELOG, and included them in the documentation via `.. include::`.

The underlying proposal is to make this "about" folder (with includes from standard files at project root) a recommendation we promote here at writethedocs.


ericholscher:
Looks like this merge conflicts with the other one (it has a bunch of similar renaming of files).

ericholscher:
Think I got it merged.

PR #10, opened 2013-01-28
fix typo

RichMorin:
On the main site, you use the term documentarian, so this should match.


thoward:
Nice catch! Thanks for the PR, @RichMorin. Looking forward to your talk proposal. :)

PR #13, opened 2013-01-31
Convert logo from image to css/javascript

thoward:


thoward:
Self-merge take two!!

PR #16, opened 2013-02-14
Upgraded sphinxcontrib-testbuild to bugfix version 0.1.2.

benoitbryon:
Version 0.1.2 of sphinxcontrib-testbuild is a bugfix release.
See https://github.com/benoitbryon/sphinxcontrib-testbuild/issues/1


benoitbryon:
To upgrade an existing install: `make update`

PR #47, opened 2015-06-29
Retire the 2015 conference callout

kochb:
It's been a month, might be time to say goodbye to the reminder.

I've reverted it to the message it had previously.


ericholscher:
Thanks! Should probably add a callout for our EU event, but as you can see, we're not great at keeping it up to date :(

PR #50, opened 2015-12-25
Update Sphinx URL

samwilson:
This change updates the old Sphinx URL http://sphinx.pocoo.org/ to the new one: http://sphinx-doc.org/. The old one redirects to the new anyway, but it seems worth having the right one.


ericholscher:
Thanks! :)

PR #73, opened 2016-08-23
Add schedule for EU 2016

ericholscher:


ericholscher:
The speaker links are currently breaking for names with unicode, because unicode. Otherwise this should be good to go once we have confirmation from folks.

PR #85, opened 2016-09-18
Move speaker data for NA 2015 into the data folder

zerok:
This should be the first step into moving talk-deta into files we can process more easily.


ericholscher:
:+1:

Issue #60, opened 2016-07-27
Add photos of our wonderful community to the front page of the site


ericholscher:
It's silly our front page doesn't have a group photo, at least.

plaindocs:
The conf front page? Or the main index?

ericholscher:
Probably both -- for being a community, we sure don't have many photos _of_ our community :)

plaindocs:
Fixed in #90

PR #88, opened 2016-09-22
Add check for template and link names in each locale.

ericholscher:
Thoughts @plaindocs?

### DIFF
--- a/docs/_scripts/test-branding.py
+++ b/docs/_scripts/test-branding.py
@@ -3,7 +3,10 @@
 import os
 import sys
 from os.path import dirname, abspath
-d = dirname(dirname(abspath(__file__)))
+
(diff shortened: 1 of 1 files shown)


plaindocs:
You're checking that we're not using `na` in `eu` and vice versa. Might be more sane to check for `na` in `na`? Ie, I added a `de` template in `eu` and the test passes.

plaindocs:
Ignore the reference to #60 , mistaken link.

ericholscher:
Yea -- it works for tito links and other stuff too this way. It's more about catching copypasta fouls than confirming correctness.

ericholscher:
There's a bunch more we could be doing -- actual template linting/confirmation could probably be better done at the Sphinx level.

plaindocs:
I will do research into the linter thing. But `time`.

PR #113, opened 2017-01-06
Clarify "independent" ticket description

brittag:
I suggest including governments in the "independent" ticket description. :)


ericholscher:
Makes sense. Thanks :)

PR #119, opened 2017-01-10
Update code-of-conduct.md

nbkaplan:
1. Used semicolons to separate the parts of the “Harassment” paragraph instead of all commas, but just because the first item has a lot of commas in it. Style issue, not critical. 

2. Can there be a designated harrassment report # that is tracked during the conference? Easier than asking someone to pick one of the 3 below and hoping they're available


ericholscher:
Closing this, as we are drafting a new COC in #122. Thanks for the contribution, though!

PR #127, opened 2017-01-22
Created 2017 Prague website and related files

thatdocslady:
First draft, tested with livehtml. Generic content was copied from the NA site, some new content was added or edited to adapt to Prague.

Still missing:

- EU logo (need to ask the designers for the svg or try to copy and edit the NA svg, I failed at it)
- CFP information (intentionally until we open it, I was thinking to copy it from NA when the time comes so I didn't bother hiding it with comments and just deleted it)
- New spreadsheet for unconference schedule
- Welcome wagon bios (I have the people confirmed but didn't get a chance to do this yet, will handle as a part of the welcome wagon trello card later and copy the format from NA)

Open issues:

- The NA schedule assumes starting at 8am, so I changed everything to be 1 hour later, but with 1.5hr for lunch it means we finish quite late. Is the 1.5hr for lunch something that we need to apply to Prague? We have the deal with autoklub to have 2 lunch stations on 2 floors, so I'd imagine that we can do with a 1hr lunch. 
- Regardless of whether we have 1hr or 1.5hr for lunch, I recommend moving the party start to 20:00, less of a chance of folks wanting dinner.
- The venue pictures in both NA and EU are not centered, tried to add an align parameter and it didn't work so maybe someone has a brilliant idea on how to fix it?


plaindocs:
We're under strict instructions not to look at this yet. :-p

thatdocslady:
Ok now the site is live-tested as much as possible, tweaked some things and updated the PR.

@plaindocs @ericholscher please review, and Eric if you can get us a staging site link for the review it'll be amazeballs :)

docs/conf/eu/2017/lightning-talks.rst
```diff
@@ -0,0 +1,123 @@
...
+
+.. _speaker-lightning-talks:
+
+Lightning Talks
+~~~~~~~~~~~~~~~
+
+A lightning talk is a very short talk where you share an idea, concept,
+or a bit of information you find interesting. They’re quick, easy, and a
+great way to practice. (Lovely example
+`*here* <https://www.youtube.com/watch?feature=player_embedded&v=6wcP1aMl7wQ#t=942>`__!)
```
plaindocs, line 11: I find linking *here* a little odd. How about Here is a `lovely example <...>`_

docs/conf/eu/2017/schedule.rst
```diff
@@ -0,0 +1,88 @@
...
+things documentation.
+
+Main Stage
+~~~~~~~~~~
+
+.. datatemplate::
+   :source: /_data/eu-2017-day-1.json
+   :template: include/schedule.rst
+
+Unconference Room
```
plaindocs, line 52: I suspect the headings got a bit muddles here.

# Schedule

## Sat

## Sun

## Mon

### Main

But the unconf is ## again, right? And I expect that will cascade down through the schedule?

docs/conf/eu/2017/unconference.rst
```diff
@@ -0,0 +1,69 @@
+:template: 2017/eu-content.html
+
+Unconference
+============
+
+As part of last year's conference, we held an
+`unconference <http://en.wikipedia.org/wiki/Unconference>`__ across the main stage room.
```
plaindocs, line 7: Should possibly be a single underscore? But looks like it works so it is probably correct.

docs/conf/eu/2017/visiting.rst
```diff
@@ -0,0 +1,140 @@
...
+flights to/from New York (JFK) and several other US airports.
+
+Flights
+~~~~~~~
+
+Prague's `Vaclav Havel airport <http://www.prg.aero/en/>`__ provides
+many direct and low-cost flights with airlines such as Czech Airlines,
+Ryanair, Wizz Air, Air Berlin, Norwegian... You get the idea. To get
+from the airport to the city center, you can either take a bus (average
+joureny time 30 minutes, cost ~30czk) or a taxi (average journey time
```
plaindocs: journey

docs/conf/eu/2017/visiting.rst
```diff
@@ -0,0 +1,140 @@
...
+If you're flying into Prague Airport (PRG), you can take bus number 100,
+119, or the Airport Express (AE) bus. Depending on where you are
+staying, these bus lines will connect you to several major metro
+stations as well as the main train station. `More information on airport
+transfers. <http://www.dpp.cz/en/public-transit-to-prague-airport/>`__
+
+Taxi companies
+~~~~~~~~~~~~~~
+
+Prague is serviced by **Uber**, and it's the most reliable and cheapest way to secure
```
plaindocs, line 79: We possibly don't want to recommend Uber?

plaindocs (approved):
Tiny nitpicks aside, this looks great! Good work! Now i'm off to review the other thing you mentioned, which I actually thought was this...

ericholscher:
Looks good, other than having the wrong logo :)

ericholscher:
Pushed the logo changes.

docs/conf/eu/2017/visiting.rst
ericholscher: Seems ok. I like Uber when abroad, as it tends to be safer and less prone to rip offs than taxis. It's the least worst option in this case, I think.

ericholscher:
I think this is probably ready to go if we want to ship it. We could hide the schedule info if we wanted to, since it's so far out, and we can fix it before publishing. I err on the side of hiding it until we can fix it. I don't think we need a 1.5hr lunch -- NA is just bigger and needs it now.

plaindocs:
logo schmogo.

docs/conf/eu/2017/visiting.rst
plaindocs: Yeah, I'm talking from a political perspective not from an end user service perspective. I suspect it will be most people's goto anyway, so we don't need to actually endorse them.

thatdocslady:
@ericholscher I merged this PR and then the whole WTD site went down (400 error), so I reverted it in PR #144 and now the site is back but it means that something needs to be fixed in this PR but I don't know what :(

PR #131, opened 2017-01-31
Fix date format in blog post

seancolsen:
Before applying this small fix I was not able to run `make livehtml`. On trying, I would receive the following error message: 

```
ValueError: invalid post date in: blog/newsletter-september-2016
```

After making this tiny edit, I was then able to run `make livehtml` successfully.


plaindocs:
Interesting, don't think I had the same error. Thanks for the fix!

PR #132, opened 2017-01-31
Documentation Principles

seancolsen:
Here is the addition of some new content to help software projects, and especially developers, follow some high-level, overarching principles when writing content and designing documentation systems. Each principle contained herein has stemmed from my annoyance at various shortcomings within software documentation. My hope is that, with WTD syndicating this set of principles, documentarians like me will be able to use them to help improve doc culture within software projects. 

I [discussed](https://writethedocs.slack.com/archives/general/p1485887060002629) this new content with @ericholscher who agreed that this content and placement within the guide would be a "a great start". 

I initially posted this set of principles to a [personal repo](https://github.com/seanmadsen/elegant-docs) for sharing. After receiving [some](https://writethedocs.slack.com/archives/general/p1485881433002622) [feedback](https://writethedocs.slack.com/archives/general/p1485892080002648) about the TOC and glossary, I chose to combine these sections into one integrated TOC/glossary section which is a little simpler. 

When/if this PR is merged, I will update the document within my personal repo to remove the content and instead link directly to WTD content.

### DIFF
--- a/docs/guide/writing/docs-principles.rst
+++ b/docs/guide/writing/docs-principles.rst
@@ -1,0 +1,367 @@
+Documentation Principles
+========================
+
+Software development benefits from `philosophies`_ and `principles`_ such as
+`DRY`_, `KISS`_, `code reuse`_, and many more. With these commonly understood
+and accepted standards, developers can hold themselves and each other
+accountable to producing high-quality code.
+
+.. _philosophies: https://en.wikipedia.org/wiki/Category:Software_development_philosophies
+.. _principles: https://en.wikipedia.org/wiki/Category:Programming_principles
+.. _DRY: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself
+.. _Don't Repeat Yourself: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself
+.. _KISS: https://en.wikipedia.org/wiki/KISS_principle
+.. _Code Reuse: https://en.wikipedia.org/wiki/Code_reuse
+
+This set of principles seeks to define similar standards for software
+*documentation* that, when practiced, will foster clean and intuitive
+content. The end goal is ultimately to delight and empower readers by
+making information easier for them to acquire.
+
+Overview, by component
+----------------------
+
+All documentation
+~~~~~~~~~~~~~~~~~
+
+In general, documentation should be...
+
+* `Precursory <#precursory>`__
+* `Participatory <#participatory>`__
+
+Content
+~~~~~~~
+
+*"Content" is the conceptual information within documentation.*
+
+**Content** should be...
+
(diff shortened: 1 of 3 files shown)


docs/guide/writing/docs-principles.rst
```diff
@@ -0,0 +1,367 @@
...
+All **sources** should be...
+
+* `Nearby <#nearby>`__
+* `Unique <#unique>`__
+
+
+Publications
+~~~~~~~~~~~~
+
+*A "publication" refers to a single, cohesive tool that readers use to consume
```
plaindocs, line 63: "Tool" might be a little off here?

plaindocs (commented):
This looks like great stuff.

docs/guide/writing/docs-principles.rst
seancolsen: Agreed. I struggled with this word choice a bit, since I wanted it to encompass a wide array of "things" that the reader uses. Do you have a better word to suggest?

plaindocs: Yeah, off the top of my head I didn't. :-p

* "entity"? Although that feels like a bit of a cop out. 
* "document" also does not really help
* 

I also think "use to consume" is misleading, something more direct such as

> A "publication" refers to a single, cohesive document that readers consume (or consult?) for a set purpose. 

Although we hit murky waters when including in application help for example.

Thoughts?

seancolsen: **"tool"/"document":** I have a slight preference for "tool" over "document" because I feel it's more general and thus applies to more forms of publications. But I don't feel too strongly and am okay changing it to "document" if that change will help get this PR merged. 

 **"consume"/"consult"** I feel pretty neutral about this change. Would you like me to change "consume documentation" to "consult for a set purpose"?

Overall, I think that having this definition is useful for distinguishing between "sources" and "publications", but that it's likely to always be a somewhat awkward definition, which is why I provided examples to make it clear what a "publication" really is. 

I want to be flexible in order to get this merged. Happy to make whatever changes that you feel are necessary.

plaindocs: I'm happy to merge as is, just giving @ericholscher and anyone else who wants to weigh in a chance to comment also.

So really I'm just brainstorming better alternatives here. In my mind, the "tool" is how you get the content, not the "content". So the browser, the pdf reader, etc. In *that* context you use the "tool" to consume the "content".

If you're thinking of the "tool" as the content, then I'd definitely lean more towards "consume" than "use to consume" (possibly this distinction is more relevant than "consume"/"consult"? I do like the focus on purpose rather than the generic "consume" though.

ericholscher (approved):
Looks good to me. Definitely a solid addition.

PR #133, opened 2017-02-01
Coc revisions

Bradamant3:
If I need to fix the branch, let me know. Submitting the PR without automatic merge so folks can take a look, since I'm later than I said I'd be with it.


mjang:
Closing to prevent an accidental messy merge

PR #135, opened 2017-02-05
Add sponsors to the main page

plaindocs:
After the most protracted battle with everything in recent memory, I think this is workable, and looks not totally ugly.


ericholscher:
Yep, I hate grids. Looks good tho.

plaindocs:
hah, I need to fix up the alignment of the RTD logo - but that is picky

PR #136, opened 2017-02-08
RTD positioning tweak

plaindocs:
Looks slightly better.

Also related, @ericholscher any thoughts or feelings on me adding Plaindocs as sponsor in kind - more than happy not to, but would if nobody objects.


ericholscher:
Seems fine to me.

plaindocs:
Done.

PR #137, opened 2017-02-09
Create newsletter-february-2017.rst

kelly-o:


hillaryfraley:
Found a few things during proofing:

- The genral consensus [general]
- Per usualy [usual]
- Then get them involve in [involved]
- gap more smoothlywas to [smoothly was]

PR #139, opened 2017-02-09
Add basic Slack Social Rules

ericholscher:
This is somewhat inspired by the Recurse Center,
but also talks a bit from my past history.

Additions/suggestions welcome.

### DIFF
--- a/docs/slack.rst
+++ b/docs/slack.rst
@@ -62,10 +62,10 @@ This concept is inspired by the `Recurse Center's Social Rules <https://www.recu
 We have a few guidelines that we hope people will follow when interacting inside out Slack network:
 
 * **You don't have to be in every conversation** - Just because you can make a comment, doesn't mean you should. Try and be constructive in all that you do.
-* **If you don't have anything nice to say, don't say anything at all** - Once again our parents were quite sage
-* **When people say hi, say hi back** - The world is a better place with a flood of hello's instead of a dearth
+* **If you don't have anything nice to say, don't say anything at all** - Once again our parents were quite sage.
+* **When people say hi, say hi back** - The world is a better place with a flood of hello's instead of a dearth.
 * **Don't be snarky** - We are all trying the best we can
-* **Be careful with jokes** - You know that you're joking, but some people might take you seriously
+* **Be careful with jokes** - You know that you're joking, but some people might take you seriously.
 * **No Black Sheep** - There may be tools that you don't like. There is a good chance there is someone in the room who either helped create or likes to use that tool. Don't make them feel bad, they are likely just in a different context.
 
 Programmer Community Issues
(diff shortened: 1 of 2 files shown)


docs/slack.rst
```diff
@@ -46,3 +48,45 @@ Chances are we have a large group of folks in your neck of the woods,
...
+
+Social Rules
+------------
+
+Our goal with Write the Docs is to be **the most welcoming community on the internet**.
+Documentarians tend to be empathetic people,
+this means that we should be good at communicating well.
+People who are new to our community in particular should be assisted and welcoming.
+
+This concept is inspired by the `Recurse Center's Social Rules <https://www.recurse.com/manual#social-rules>`_ -- which you should also follow.
```
plaindocs: I find the 

> which you should also follow 

kinda confusing. Surely better to take enough from there so people don't have to read both. Also OMG long!

docs/slack.rst
```diff
@@ -46,3 +48,45 @@ Chances are we have a large group of folks in your neck of the woods,
...
+
+We have a few guidelines that we hope people will follow when interacting inside out Slack network:
+
+- **You don't have to be in every conversation** - Just because you can make a comment, doesn't mean you should. Try and be constructive in all that you do.
```
plaindocs: Would this be better worded as something like 

> **Only participate when you can be constructive** - ...?

docs/slack.rst
```diff
@@ -46,3 +48,45 @@ Chances are we have a large group of folks in your neck of the woods,
...
+- **Say hi** - The world is a better place with a flood of hello's instead of a dearth.
```
plaindocs: might need a dearth of greetings?

docs/slack.rst
```diff
@@ -46,3 +48,45 @@ Chances are we have a large group of folks in your neck of the woods,
...
+and I'd like to make sure we don't fall into these same traps.
+
+In particular,
+please avoid these behaviors:
+
+- **No Black Sheep** - There may be tools that you don't like. There is a good chance there is someone in the room who either helped create or likes to use that tool. Don't make them feel bad, they are likely just in a different context.
+- **Let Me Google That For You (LMGTFY)** - If people ask questions that are simple to answer on google, still answer them nicely.
+- **Question Policing** - Some people might ask incomplete questions that are hard to answer. Ask follow up questions and act in good faith to help them solve their issues.
+- **Technology Pissing Contests** - We don't care if your tool is faster, better, bigger, stronger. We care about creating great documentation for our users.
+
```
plaindocs: Most of these fall under 'assume the best of people' which might be worth stating?

plaindocs: Do we need to explicitly add a no bashing lang / os / editor or do we think it is sufficiently covered by Technology Pissing.

Also, 

> faster better bigger strong tool?

docs/slack.rst
```diff
@@ -46,3 +48,45 @@ Chances are we have a large group of folks in your neck of the woods,
...
+Reporting Issues
+~~~~~~~~~~~~~~~~
+
+If you see people not following our etiquette guidelines,
+you have two options:
+
+* Message them privately, and explain why what they said might have been hurtful to someone
```
plaindocs: > referring to this guide

plaindocs (commented):
Good basics.

plaindocs:
Good to merge for now I think. :+1: :+1: :+1:

PR #140, opened 2017-02-12
Remove CFP

plaindocs:
Does what it says.


plaindocs:
I just someone to sanity check, and reckon it is time to not bug Eric on every PR. :-p

PR #141, opened 2017-02-13
Add a page about automated testing

ericholscher:
This is a quick brain dump,
but seems like it should exist.


thatdocslady:
I like this, can we merge and then maybe flag for next month's newsletter. In fact maybe we should have a dedicated "what happened in Write the Docs this month" section, where things like new blog posts, conf updates, meetups, etc, can be groups instead of submitted separately? cc @kelly-o :)

svx:
I also would add that for Sphinx is makes sense to run test-builds in nit-picky mode, that is for catching build errors and sphinx-rst validations.

Some other basic ci test, could/should be a test where you check your docs against word consistency, to make sure that you always use the same wording like for example, always use JavaScript through the whole docs and not sometimes Javascript.

These are basic tests, but they already help a lot.

Maybe it would be also nice to add/link to some examples ?
There are some in the [training](https://github.com/testthedocs/ttd.training-docs-devs)  we did for devs and some in the [one](https://github.com/testthedocs/ttd.training-docs-basics) for everyone.

This are 'just' really basic stuff, I still have to publish all the 'cool' stuff :)

plaindocs:
I've mentioned picky mode. 

Examples would be nice, but I'd prefer them to be targeted and explained not just huge repos of stuff.

I'm merging this for now, but more comment welcome any time.

PR #151, opened 2017-02-21
Resize logo

plaindocs:
That looks like it kinda works?


ericholscher:
SVG's are scalable -- and should scale to any size. Unless you're changing the aspect ratio, it shouldn't be necessary?

plaindocs:
Yeah, I had to recrop it as one was cropped with padding.

PR #152, opened 2017-02-21
Replace easy_install with pip

astrojuanlu:
Nowadays `easy_install` should not be recommended anymore as the primary Python package manager so I replaced it by `pip`. Although running `sudo pip install` is potentially dangerous, can break the system in hard to debug ways and in general should never be used instead of the system package manager, I understand that you want something _that works_ so I'm leaving it as it is too. Anyway, expanding this section a little bit with some warning would be also desirable IMO.


ericholscher:
:+1: Thanks!

PR #156, opened 2017-02-22
Missing verb in section header

baggerspion:
Added the missing verb to the section header so it makes a little more sense :)


ericholscher:
Thanks :)

plaindocs:
Verb the docs!

PR #170, opened 2017-03-15
[WIP] Break JSON up into specific types

ericholscher:
This is the first step in cleaning up our JSON rendering. 
It also starts storing Markdown and then rendering HTML,
so that the abstracts are easier to change and more consistent.


ericholscher:
Need to do some more thought about how to integrate this into the HTML templates. It is non-trivial >:x

plaindocs:
Probably we want to reduce the size of the template to just the table rows, and leave the 3 headings in the RST?

That sound reasonable?

plaindocs:
@ericholscher I can take a look if you're busy - lemme know.

ericholscher:
Sure -- I'm probably not gonna have much time to poke at this for a few more work days.

plaindocs:
FIXED ~Not sure where the empty `UL` at the start of each section is coming from.~

What does [html_context](http://www.sphinx-doc.org/en/stable/config.html#confval-html_context) do? Just provide that variable to the html renderer? Which is why we have to add the speaker arrays?

DONE ~We should probably at least have a link to each section at the top of the page.~

plaindocs:
We need to fix up the CSS:

```
<div class="contents local topic" id="contents">
<ul class="simple">
<li><a class="reference internal" href="#workshops" id="id1">Workshops</a></li>
<li><a class="reference internal" href="#talks" id="id2">Talks</a></li>
<li><a class="reference internal" href="#unconference-sessions" id="id3">Unconference sessions</a></li>
</ul>
```

plaindocs:
OK, update on this. I've futzed around a little, added the text @kelly-o mentioned. The blue links look ok.

There is some ugly display stuff that I can't easily fix though. We could remove the mini TOC but meh.

minitoc

ericholscher:
Yea, I got that far as well, and the design exploded, which is why I stopped too.

plaindocs:
Right. I'll rethink it and have another bash.

plaindocs:
I'm thinking we can close this, put up with it till next redesign.

PR #171, opened 2017-03-20
Most of these words are not wanted in tech docs. WIP

plaindocs:
THIS IS ON HOLD AT THE MOMENT

Of course, life is not that simple, and we do have blog content as well as tech docs, but we can track that list here for the moment anyway. 

PRs to this branch with new words welcome, discussion on alternative checks (such as words you can suggest an alternative to instead of dropping) in the comments below.

### DIFF
--- a/vale/WTD/Just.yml
+++ b/vale/WTD/Just.yml
@@ -1,0 +1,10 @@
+extends: existence
+message: "Consider removing '%s'"
+ignorecase: true
+scope: text
+level: warning
+tokens:
+  - just
+  - easy
+  - simple
+  - obviously
--- a/vale/WTD/Just.yml
+++ b/vale/WTD/Just.yml
@@ -4,6 +4,7 @@ ignorecase: true
 scope: text
 level: warning
 tokens:
+  - awesome
   - clearly
   - easily
   - easy
--- a/vale/WTD/Just.yml
+++ b/vale/WTD/Just.yml
@@ -4,7 +4,15 @@ ignorecase: true
 scope: text
 level: warning
 tokens:
-  - just
+  - clearly
+  - easily
   - easy
-  - simple
+  - essentially
+  - generally
+  - in order to
+  - just
+  - obvious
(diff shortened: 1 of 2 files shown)


mjang:
Should we include the weasel words cited in write-good -- including those listed in this shell script? http://matt.might.net/articles/shell-scripts-for-passive-voice-weasel-words-duplicates/

weasels="many|various|very|fairly|several|extremely\
|exceedingly|quite|remarkably|few|surprisingly\
|mostly|largely|huge|tiny|((are|is) a number)\
|excellent|interestingly|significantly\
|substantially|clearly|vast|relatively|completely"

plaindocs:
Yes! 

Vale already has a lot of the `write-good` stuff built in (https://github.com/ValeLint/vale/tree/master/styles/write-good) , but perhaps not weaselwords. Either way, I'll add a PR to Vale for those. Cheers for the suggestion.

One of the things I want to do with this whole thing is either find a standard that different linters can use, or a translator for source files so people aren't constantly reinventing the wheel.

mjang:
Gosh, Vale looks awesome -- especially as it's lists are in Yaml files (which means customization is straighforward).

vale/WTD/Just.yml
```diff
@@ -0,0 +1,19 @@
+extends: existence
+message: "Consider removing '%s'"
+ignorecase: true
+scope: text
+level: warning
+tokens:
+  - awesome
```
ericholscher, line 7: I like awesome -- feels like it should be within out brand voice :)

plaindocs: Yeah, you've reminded of something I've already realized. A lot of WTD content is more informal blog than docs. I'm tracking here anyway, and will split before merging.

ericholscher: Yea -- I imagine we can apply it to specific directories, for example the guide and the conf websites should probably have different voice/tones.

PR #172, opened 2017-03-22
This reads better.

plaindocs:
Manual fixup so we're not talking about morning workshop and workshop 2.

CC @kelly-o


ericholscher:
:+1:

PR #173, opened 2017-03-22
Reload on json changes

plaindocs:
I don't *actually* think this works, but it would be nice if it did.


ericholscher:
Why doesn't it work? Seems like it should.

plaindocs:
Dunno. Something to do with how the templates are interpreted? But that is an out and out guess.

Issue #159, opened 2017-02-27
Cleaning Undefined label warnings


plaindocs:
We get a TON of these on clean builds:

```
/_data/na-2017-day-1.json:1: WARNING: undefined label: 15-minute-talk (if the link has no caption the label must precede a section header)
/_data/na-2017-day-1.json:1: WARNING: undefined label: 5-min-switch-speakers (if the link has no caption the label must precede a section header)
/_data/na-2017-day-1.json:1: WARNING: undefined label: 15-minute-talk (if the link has no caption the label must precede a section header)
```

are they an unavoidable consequence of json/markdown? or can I fix by specifying Titles in either links or anchors? 

Also, these. Do we have to add years to links?

```
/home/samuel/Nosync/Repositories/WTD/www/docs/conf/eu/2016/index.rst:17: WARNING: Duplicate explicit target name: "videos".

```

plaindocs:
I'm doing some background cleanup work in https://github.com/writethedocs/www/tree/sw-cleaning

ericholscher:
Yea, that was from my janky hacks that automated linking from the schedule to the talk abstract. +1 to cleaning them up or doing it in a smarter, more data driven way :)

plaindocs:
Running a build with sphinx 1.5.3 not the 1.3.6 in `requirements` also spits out an error related to awesome font:

```
WARNING: The config value `fontawesome_link_cdn' has type `str', defaults to `bool'.
```

ericholscher:
Yea, I think that might be a bug with `ablog`. I should investigate more, but I don't think it's a real issue, as the config says it should be a string. http://ablog.readthedocs.io/manual/ablog-configuration-options/#confval-fontawesome_link_cdn

plaindocs:
Marking this as done.

PR #176, opened 2017-03-23
More cleanup WIP

plaindocs:
This is a not very pretty fix for most of the ugly warnings on the 2016 confs #159

### DIFF
--- a/docs/_templates/include/schedule.rst
+++ b/docs/_templates/include/schedule.rst
@@ -9,7 +9,7 @@
         <div class="col-xs-10>
           <p class=">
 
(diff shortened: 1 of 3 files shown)


ericholscher:
Ha, that's one way to do it :)

ericholscher (approved):

plaindocs:
Unless I'm missing something the ref syntax doesn't work if we're generating html not rst.

I'll merge this, and if someone cares enough they can suggest a better solution. :-p

ericholscher:
Hrm, this ended up linking all the content like "Snack Break" and "Lunch" to non-existing entries in the speaker page, which seems like a subpar outcome. We should probably just special case the logic to not apply to those specific titles.

PR #177, opened 2017-03-24
Add some information about test-driven API documentation

wilkinsona:
Test-driven API documentation is, as far as I know, a new approach to producing API documentation. I've seen some interest from member of the Write The Docs community so hopefully it's something that you'd like to include on your website.

If there's interest in a more in-depth discussion of the approach, I'd be more than happy to make another contribution.

### DIFF
--- a/docs/guide/api/about-api-documentation.rst
+++ b/docs/guide/api/about-api-documentation.rst
@@ -22,6 +22,13 @@ While these tools help you with the technical details, they don't create good de
 
 .. _documentation-driven-design:
 
+Test-Driven Documentation
+-------------------------
+
+Test-driven documentation aims to improve upon the typical approaches to automated documentation. It allows you to write the bulk of the documentation by hand while also ensuring its accuracy by using your API's tests to generate some of the content.
+
+Projects such as `Spring REST Docs <https://spring.io/projects/spring-restdocs>`_ use your API's tests to generate small snippets of documentation that can be included in hand-written API documentation. The accuracy of the documentation is ensured by the tests – if the API's documentation becomes inconsistent with its implementation the tests that generate the snippets will fail.
+
 Documentation Driven Design
 ---------------------------
 


docs/guide/api/about-api-documentation.rst
```diff
@@ -22,6 +22,13 @@ While these tools help you with the technical details, they don't create good de
 
 .. _documentation-driven-design:
 
+Test-Driven Documentation
+-------------------------
+
+Test-driven documentation aims to improve upon the typical approaches to automated documentation. It allows you to write the bulk of the documentation by hand while also ensuring its accuracy by using your API's tests to generate some of the content.
+
+Projects such as `Spring REST Docs <https://spring.io/projects/spring-restdocs>`_ use your API's tests to generate small snippets of documentation that can be included in hand-written API documentation. The accuracy of the snippets is guaranteed by using your API's tests to generate them – if the snippets become inaccurate, the API's tests will fail.
+
```
plaindocs: This feels every so slightly circular to me. Use API tests to generate snippets, if the snippets are inaccurate the tests fail. 

Is there something I'm missing in the wording?

plaindocs:
My one comment aside this looks great - happy to merge in as soon as we address that - and if you want to write more that is definitely welcome.

docs/guide/api/about-api-documentation.rst
wilkinsona: Thanks for the review. I don't think you've missed something as the wording was ambiguous. I've force-pushed a new commit that hopefully makes things clearer

wilkinsona:
> My one comment aside this looks great - happy to merge in as soon as we address that

Thanks

> and if you want to write more that is definitely welcome

Excellent. Where do you think it should go? Would a new top-level entry within the API Documentation section be an appropriate place for a longer discussion of the approach?

plaindocs:
Do you think that `test-driven` is separate from the stub at `http://www.writethedocs.org/guide/api/testing/`? That seems like the best place for me. We can always take another look when you've put some content in there.

plaindocs:
Thanks! :+1:

wilkinsona:
The Testing section looks like a good place to me.

PR #179, opened 2017-03-25
Revise JSON for Meetup Unconf Description

mjang:
I just realized that we have the wrong description for our Meetup unconference. (we're not doing a Microcopy talk at the unconf ;))

I'm not sure how to test the output for this, so I need a review.


plaindocs:
Looks good to me.

If you can be bothered to install the sphinx etc tool chain instructions in the README should work, and you can preview with `make livehtml`. 

If they *don't* work I'd love to hear about it! :smiling_imp:

mjang:
Thanks @plaindocs, it does work 
screen shot 2017-03-26 at 7 39 12 am


(mulling the README -- while "as the root user" is direct, it seems awkward w/r/t security)

ericholscher:
You can use virtualenv instead of running as root, and should be the recommended action, but adds complexity.

plaindocs:
Thanks for testing @mjang 

@ericholscher I'm happy to rewrite using `virtualenv` but I agree on the complexity - and as always I'm reluctant to raise the bar to submitting PRs

PR #180, opened 2017-03-25
Add Laura to the meetup unconf

ericholscher:
This I believe is the third person who wanted to be involved in the unconf.


mjang:
@meker12 I think we need to do this, but I'm going to trust your judgement on this one. Should we add Laura to the list for the Meetup Unconf?

meker12 (approved):
Looks good to me. We'll put together an agenda to balance out the unconference.

PR #182, opened 2017-03-27
Add sponsors

plaindocs:
@ericholscher 
Can you sanity check this for me?

Do we need to leave the mediums on their own row, and only have the small in multi column? Ah, I need to fix up.


plaindocs:
Still not perfect. I wonder if we could setup sponsorship levels in CSS?

ericholscher:
No, they are paying us a good deal less, and should have a small logo. Guessing the prospectus is wrong there, but we should just make it a small logo.

ericholscher:
Also dislike how strongly the sponsors are shown on the front page. We should probably add more content there, so that it doesn't just feel like a brochure for them.

plaindocs:
I'm less bothered by the front page, as it is going to be past the scroll for most people (I think) so... But we can add some more newsy item blocks.

PR #183, opened 2017-03-28
Fix spelling mistake in vale configuration

dsyer:


plaindocs:
Awesome thanks!

if you feel like sending the fix back upstream the same change could be made here: https://github.com/ValeLint/vale

dsyer:
https://github.com/ValeLint/vale/pull/35

PR #185, opened 2017-03-30
Created workshop announcement and updated index page

thatdocslady:
I know it's kind of a slim announcement, but honestly the two important things there are "buy tickets for this" and "read all the info on the website", no?

### DIFF
--- a/docs/conf/na/2017/index.rst
+++ b/docs/conf/na/2017/index.rst
@@ -94,7 +94,7 @@
                     <div class="section__header subheader">
                       <span class="subheader__yellow"></span>
                       <h2 class="subheader__header">Latest News</h2>
-                      <a href="news.html" class="subheader__more">Read all news</a>
+                      <a href="news/" class="subheader__more">Read all news</a>
                     </div>
                   </div>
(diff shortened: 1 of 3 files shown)


ericholscher:
We mentioned the workshops in the last announcement, and IIRC we never sent it to the main email list (http://www.writethedocs.org/conf/na/2017/news/announcing-presentations/) -- so might be worth sending that as well?

plaindocs (approved):

thatdocslady:
@ericholscher I'm still waiting for the testing on the mailchimp theme so that we can update the global template and then I can author and send an NA announcement with all those things. I wanted to have a blog post set up on the website so that we can link to it (also when I send a message via Tito to people who already got tickets).

thatdocslady:
Hmm. Now that I think about it a bit more, maybe instead of creating a separate announcement for the workshops I can back-change the previous blog post, and then link from there to the workshops page (and then link to *that* from the mailchimp/tito messages)?

plaindocs:
I don't think the extra news item is a bad thing. I'd leave it as is.

PR #188, opened 2017-04-01
Update welcome-wagon .rst

ondine:
Second time's the charm? I'm not sure where the speakers link is getting its reference. See screenshot below.

screenshot 2017-03-31 20 46 24


plaindocs:
Ah yes, that would do it! Thanks!

We should really be using the `:doc:/conf/na/2017/speakers` syntax throughout, but varying levels of familiarity with rST.

PR #193, opened 2017-04-05
added MAX construction detour information

dblando:
Eric Holscher catted with me on Slack. He suggested that I take the notes I placed into the wtd-conferences channel and add them to this document.

I replaced the content from the last sentence in `The MAX` § through the entire `The Streetcar and Buses` §. There are links to the construction notice from TriMax along with alternate paths.


plaindocs:
Awesome stuff! Thanks!

PR #195, opened 2017-04-13
Break out the Sunday schedule from Writing Day & Workshops

ericholscher:


plaindocs:
Late to this party but looks good to me.

PR #196, opened 2017-04-15
Delink non talks

plaindocs:
This is an interim fix for broken links.

We should probably look at tweaking the json / template so we can include arbitrary links for workshops and writing day.

IF link use that - if not use speaker link type thing.


plaindocs:
@ericholscher I'm merging this - shout if anything looks odd :-p

PR #202, opened 2017-04-25
Clarify schedule quickfix

plaindocs:
* [ ] No idea what is up with these original heading levels -> they get *eaten*


ericholscher:
Merged :+1:

PR #203, opened 2017-04-26
Child care option

plaindocs:
Basic. Any improvements to suggest?

### DIFF
--- a/docs/conf/eu/2017/sponsorship/prospectus.rst
+++ b/docs/conf/eu/2017/sponsorship/prospectus.rst
@@ -189,6 +189,12 @@ Sponsorship of the official party includes:
 - Mention of company on schedule for the party, and an on-stage thanks
 - Your logo on a banner or roll up at the party venue (provided by sponsor)
 
+Child Care
+-----------
+
+Many of our attendees have children, sponsoring day care makes it  easier for
+them to attend.
+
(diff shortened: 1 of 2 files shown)


docs/conf/eu/2017/sponsorship/prospectus.rst
```diff
@@ -189,6 +189,12 @@ Sponsorship of the official party includes:
 - Mention of company on schedule for the party, and an on-stage thanks
 - Your logo on a banner or roll up at the party venue (provided by sponsor)
 
+Child Care
+-----------
+
+Many of our attendees have children, sponsoring day care makes it  easier for
```
thatdocslady, line 195: double-space action ;)

also, do we not put a price on this?

plaindocs: We'd have to spec it out if we wanted to put a price on it no?

thatdocslady: True, but maybe we can add more details like "exact sponsorship amount TBD based on demand, but would probably be in the hundreds" or something more elegant like that ;)

plaindocs: I'm all about the elegant.

PR #205, opened 2017-04-26
Ugly fix for link to wrong dates...

plaindocs:
To be refactored.


docs/conf/na/2017/schedule.rst
```diff
@@ -66,7 +66,7 @@ Main Stage
 
 .. datatemplate::
    :source: /_data/na-2017-day-1.json
-   :template: include/schedule.rst
+   :template: include/schedule2017.rst
```
ericholscher, line 69: Believe it's used in the writing day as well above ^^

plaindocs: There are no links in the writing day though, so that doesn't matter :-p

PR #208, opened 2017-05-02
Add real 2017 schedule

ericholscher:


docs/_templates/include/schedule2017.rst
```diff
@@ -32,20 +32,10 @@
         <div class="col-xs-2">
           <p>{{ talk.Time }}</p>
         </div>
-        <div class="col-xs-10>
-          <p class=">
+        <div style="color: black;" class="col-xs-10">
+          <p>
 
-          {% if talk.Session in dont_link %}
```
plaindocs, line 38: Why are we removing the links?

plaindocs:
LGTM aside from not linking to abstracts.

thatdocslady:
@plaindocs you ain't slacking today ;) I suggest that we discuss this PR in here and not in the other PR, because confusing ;)

What is the problem with the links to the abstracts and do you think this is an easy enough fix that Eric can do it today? Or that you can do it tomorrow?

plaindocs:
The nitty gritty is we switched the order of the name and the title in the json data file, and so the code that auto generates the slug which we use to make the links broke.

I'd assume one of us *will* be able to fix it tomorrow, yes.

docs/_templates/include/schedule2017.rst
ericholscher: Because we don't have the data in the JSON using this method, so they don't work anyway. We were preprocessing the JSON in the conf.py, but that isn't happening in this method.

ericholscher: We need to build a proper way to handle this, likely pre-processing the JSON before checking it into the repo.

PR #211, opened 2017-05-08
Add COC reporting people

ericholscher:


plaindocs:
I have a strong aversion to 

> to the below 

and would prefer 

> to the following people

But happy to be overruled if that is US usage or something.

thatdocslady:
+1 for "following people" vs "below", and also I will need to change my phone number to whatever US-SIM I get when I arrive in PDX but for now it's ok to have my CZ number in there.

Aaaand maybe we can add our Slack handles too? In case people want to DM us there

ericholscher:
WFM :+1:

PR #228, opened 2017-05-14
Dont break on make clean

plaindocs:
Adding the dependency stops linkcheck crashing.


plaindocs:
@hwesta

Hwesta:
Fixed my issue, thanks!

Original issue was that `make linkcheck` failed if it was run before `make html` (or after `make clean`)

annegentle:
Oh, nice! Learned something today.

PR #229, opened 2017-05-14
Moving data from JSON to YAML

zerok:
This should make editing longform-content like session abstracts much more
convenient.

This uses https://gist.github.com/zerok/0a39f63cb3ad337bf3b2e40ce2a43975


plaindocs:
<3

PR #242, opened 2017-06-02
Switch to YAML

plaindocs:
This replaces #233 and contains all of the work done by @zerok to convert the data stores to YAML.

Currently produces build warnings if you run `make json` *before* `make html`. 

The aim is to test this with an updated blank EU schedule, and merge it in asap so we can start adding video and slide tags to talks. 

Also apologies for the git mess! :-D 

Tweaked times attached:
[MainConf.txt](https://github.com/writethedocs/www/files/1047003/MainConf.txt)


plaindocs:
Ping

plaindocs:
Safe to delete `docs/meetups/index.md` I think.

docs/meetups/index.rst
```diff
@@ -0,0 +1,55 @@
...
+
+.. meetup-listing::
+    :region: Oceania
+
+
+Current Meetups: Asia
+---------------------
+
+.. meetup-listing::
+    :region: Asia
```
ericholscher, line 44: My only thought is that this seems overly complex for listing our meetups. Is there a reason to move to this format? In general we often ask folks to add their meetup to the list, and having them run into this would I think be a pretty big barrier.

docs/_data/meetups/usa-portland.yaml
```diff
@@ -0,0 +1,6 @@
+region: North America
+country: USA
+state: Oregon
+city: Portland
+notes: |
+  Organised by [@WriteTheDocsPDX](https://twitter.com/WriteTheDocsPDX)
```
ericholscher: This seems broken because there is no meetup website. We should raise an error in the extension here, I think?

docs/meetups/index.rst
plaindocs: I'm unsure. People don't have to edit the index, just add a single YAML with fairly straightforward info.

@zerok and I did discuss the added complexity. 

The reason to move towards structured meetups is that it lets us add meetup videos to them, and iterate over the same list.

docs/_data/meetups/usa-portland.yaml
plaindocs: :+1:

plaindocs: Probably we should require `Country`  / `website`

plaindocs: So I don't have access to sphinxerror in conf.py? in L137

```py
if not 'website' in meetup:
            raise SphinxError('Meetup needs a website')
```

Results in

```
Exception occurred:
  File "conf.py", line 138, in load_meetups_by_region
    raise SphinxError('Meetup needs a website')
NameError: name 'SphinxError' is not defined
```

ericholscher: Where are you expecting SphinxError to come from?

ericholscher:
I think this is pretty much ready, just some small tidbits I mentioned, but should be mergable.

docs/_data/meetups/usa-portland.yaml
plaindocs: I lazily googled errors in sphinx, and expected to have access to it within conf.py

plaindocs:
Errors are vale complaining about content in speaker abstracts, which we shouldn't really care about - safe to ignore.

PR #244, opened 2017-06-06
Add WTD.Terms to Vale configuration

jdkato:
This updates the existing Vale configuration to include a simple rule for detecting misuse of "meetup" (as described in the [WTD style guide](https://github.com/writethedocs/www/blob/master/docs/style-guide.rst#meetups)). Line 56 of [`schedule-welcome-wagon-events.md`](https://github.com/writethedocs/www/blob/master/docs/conf/na/2016/news/schedule-welcome-wagon-events.md) was the only issue found.

### DIFF
--- a/.vale
+++ b/.vale
@@ -5,7 +5,6 @@ MinAlertLevel = warning
 BasedOnStyles = WTD
 
 vale.Redundancy = YES
-#vale.Repetition = YES
 vale.GenderBias = YES
 
 TheEconomist.UnnecessaryWords = YES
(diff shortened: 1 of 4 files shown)


plaindocs:
Hey @jdkato good to see you here again.

The problem here is that the meetup you fixed isn't one of the meetups mentioned in the style guide. Ie not hosted on meetup.com 

In this instance, I think the correction still applies, but we could easily write lets meet up and write some docs and it would be correct.

plaindocs:
I've been running into a *lot* of that kind of thing, and subsequently dialed back my deep dive projects into text linters lately

jdkato:
Ah, that's a good point. I suppose something like

> Let's meet up at the next Write the Docs meetup and write the docs for our new project.

would be particularly problematic.

That said, I think having a defined writing context is often enough to overcome these nuances&mdash;but I'll go ahead and close this.

PR #254, opened 2017-07-06
Proof of concept for RSS feeds for meetup events

plaindocs:
Some *rough and ready* python gleaned from the internet to get events from meetup.com . We could certainly do this better with the API.

Currently spits out something like this

```
History of the New Relic Documentation Site, Part One
Writing Inclusively About Technology Topics
Hackup
Hackup
Hackup
Hackup
Hackup
Hackup
ATX Write the Docs monthly meeting
ATX Write the Docs monthly meeting
ATX Write the Docs monthly meeting
ATX Write the Docs monthly meeting
ATX Write the Docs monthly meeting
ATX Write the Docs monthly meeting
Finding the right work to do: Lessons learnt from a year at a startup
September theme - tbc
First Write The Docs Montreal Meetup!
```

We'd need to tweak the code to grab more info than the headlines, and throw it into a template.

Decoupled from `conf.py` so we can run it as a cron with CI. FAO @rosewms

### DIFF
--- a/docs/_scripts/event_feeds.py
+++ b/docs/_scripts/event_feeds.py
@@ -1,0 +1,49 @@
+import yaml
+import feedparser
+import glob
+import io
+from pprint import pprint
+
+def load_yaml(path):
+    with io.open(path, encoding='utf-8') as fp:
+        return yaml.safe_load(fp)
+
+def load_meetups():
+    result = []
+    for yaml_file in glob.glob('../_data/meetups/*.yaml'):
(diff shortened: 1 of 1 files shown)


ericholscher:
This is a nice outcome of having the meetups in YAML :)

Thoughts on next steps here?

plaindocs:
After taking a look at a different RSS, they have dates in the text body, which is rubbish, so I'll adapt to use the API.

PR #265, opened 2017-07-26
Add breadcrumbs to the main doc pages.

ericholscher:
screenshot 2017-07-26 12 24 18

This will help with understanding where you are,
and acts as a basic navigation aid.

### DIFF
--- a/docs/_static/css/global-customizations.css
+++ b/docs/_static/css/global-customizations.css
@@ -7,3 +7,7 @@
     border: 2px dotted;
     border-color: red;
 }
+
+.breadcrumbs {
+    margin-bottom: 2em;
+}
--- a/docs/_static/css/global-customizations.css
+++ b/docs/_static/css/global-customizations.css
@@ -8,6 +8,3 @@
     border-color: red;
 }
 
-.breadcrumbs {
-    margin-bottom: 2em;
-}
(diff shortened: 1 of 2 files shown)


plaindocs:
This doesn't seem to work like I would expect it to. It gives you the parent page, but not the current one - which I think is intentional. 

How about we include the current page name, but not linked, to indicate it is the current page.

capture image2
breadcrumbs

plaindocs:
https://www.smashingmagazine.com/2009/03/breadcrumbs-in-web-design-examples-and-best-practices/

ericholscher:
The current page name is the title of the page, directly under it. It feels redundant to have it both places, but I don't care too much.

ericholscher:
Also including the current page name sometimes causes the breadcrumb to wrap, making it look extra janky to include twice.

plaindocs:
Hmmmm, I do see your point. Esp with the wrap.

I still find it confusing to *not* see the page I'm on.

@thatdocslady @kelly-o @mjang @Bradamant3 more opinions?

ericholscher:
The big huge title at the top of the page is the page you're on, no? 

We do repeat it here: http://docs.readthedocs.io/en/latest/getting_started.html -- but it feels a bit more detached in the design.

Bradamant3:
The RTD example also omits the `>>` at the end of the breadcrumb. And I like the additional `>>` as a solution to the otherwise-repeat-the-title problem/situation. IOW -- I vote for no redundancy/repetition.

plaindocs:
Fair enough, I'm probably overthinking.

PR #281, opened 2017-08-21
Removes a sentence from the afternoon intro

sarahmaddox:
Removes "This is the main event!" from the afternoon section, because it implies that the doc sprint isn't as important/cool. :)


plaindocs:
Thanks!

PR #295, opened 2017-09-11
Add WTD EU 2015 speakers as data file

zerok:


plaindocs:
I've not checked all links and bios, but assume it is all perfect. If any mistakes did creep in, we can fix after. THANKS @zerok

PR #239, opened 2017-05-23
Static templates for Write the docs conferences and meetup videos 

xpostudio4:
I created templates to display both videos and the individual conferences videos and the meetups.

The relative urls to access the templates are:
/conf/na/2017/videos/
/conf/na/2017/conference-videos/
/conf/na/2017/individual-meetup/
/conf/na/2017/video-detail/

Here I attach the original images for the views as well as the original sketch if anyone needs reference for the proposal. If anything needs to be changed
[Write the Docs.zip](https://github.com/writethedocs/www/files/1023966/Write.the.Docs.zip)

 just let me know.

individual video view
meetups individual view
meetups view index
conference views index

PR #297, opened 2017-09-12
Video archive for the conferences

zerok:
This so far only includes video pages for the conferences using the *.*.speakers.yaml data file format. I'd propose that we handle meetups in a different PR.

This is based on #239 . Big thanks to @xpostudio4 for his layouts 😄


zerok:
The build breaks right now because it is expecting the data from #295 to be there.

plaindocs:
Awesome, I'll have a play around with this, and see if I can break it.

zerok:
Awesome, thanks :) I've merged the updated master and the build now goes through :)

ericholscher:
Got this running locally with a bit of futzing. Looks good to me!

zerok:
@ericholscher Awesome! Thank you :D

PR #298, opened 2017-09-13
Adding link to slides (plus resources and rixx's notes) to EU 2017 sp…

cory-wc:
Adding link to slides (plus resources and rixx's notes) for Cory Williamson-Cardneau talk on microaggressions


plaindocs:
Thanks! <3

PR #314, opened 2017-09-27
Work in Progress 2018 Themes

ericholscher:
Ship new website.

# High Priority (before launch)

* [x] margin-bottom on `li` is broken (https://deploy-preview-314--dentist-canary-64055.netlify.com/conf/portland/2018/visiting/#eating see above "Where to stay")
* [x] Fix schedule or remove it (Now that we have HTML, we comment out the wrong data)
* [ ] Fix news listing layout from Sphinx (https://deploy-preview-314--dentist-canary-64055.netlify.com/conf/portland/2018/news/)
* [x] link up social media icons
* [ ] Fix alignment of schedule overview in home page (titles are not on the same line)
* [x] Proofread and update text (i.e. some dates are wrong, typos, etc) 

# Low Priority (after launch)
* [ ] Remove dummy news items from the home page (currently not linking anywhere, we will create a blog post and then we can link to it)
* [ ] Change color of headings to black/dark gray (green is too light) - accessibility
* [ ] Change hyperlink colors to something darker (maybe a darker green?) - accessibility
* [ ] Make bold bolder
* [ ] Make primary text black instead of grey, for accessibility (or at least a darker gray)
* [ ] fix emphasis (the red looks pretty harsh to me, but no idea what would be better)
* [ ] The side margins of the content pages could probably be smaller (i.e. the width of the center column could be bigger)
* [ ] "Portland" text in the logo shouldn't be bold 
* [ ] Remove second instance of H1 in the content pages (currently the heading text appears both in white over an image, and in black below the image, and it should only appear once in white)
* [ ] Tickets page has incorrect color scheme for the actual ticket types

# Non design tasks

* [x] Fix `TODO`s eg in schedule
* [x] add sponsors
* [x] switch slack link to signup (slack.writethedocs.org)
* [x] change google form colour from yellow


plaindocs:
Having a click through of the preview now:

* some slightly odd spacing around https://deploy-preview-314--dentist-canary-64055.netlify.com/conf/portland/2018/visiting/#where-to-stay looks like a list/heading clash

plaindocs:
It is looking awesome!

plaindocs:
We've got `*not*` coming out as red - which is probably unintentional (in https://deploy-preview-314--dentist-canary-64055.netlify.com/conf/portland/2018/visiting/#taxi-companies)

plaindocs:
EDIT: moved all this up to the desc

plaindocs:
Also, we *really* need to get the post list in index to use this:

http://ablog.readthedocs.io/manual/posting-and-listing/#listing

I'll have a look shortly. 

EDIT:

Looks like we can just 

```
.. postlist:: 3
   :date: %A, %B %d, %Y
   :format: {title} by {author} on {date}
   :list-style: circle
   :excerpts:
```
And then format the output with 

```
<ul class="postlist-style-circle postlist">
```

EDIT2:

except we'd need to be able to do that in RST. I think we could leave everything else in the html template and include the news items as the rst body. @xpostudio4 what do you think?

ericholscher:
The trickiest thing w/ intermixing jinja templates & RST HTML is the `pathto` function, which I believe won't exist in the `.. raw:: html` bits. I need to figure out a good way to make that less janky.

xpostudio4:
Regarding that we can add the actual route when generated, it seems to work for me.

ericholscher:
Yea, but I'd like to retain the relative URL's if possible, so it builds locally across different builders. It's a pattern we need generally, so good to figure out.

xpostudio4:
@plaindocs I would really appreciate an example, I suppose you used the same process last year, right?

ericholscher:
@xpostudio4 What do you need an example of?

plaindocs:
> The trickiest thing w/ intermixing jinja templates & RST HTML is the pathto function, which I believe won't exist in the .. raw:: html bits. I need to figure out a good way to make that less janky.

From a **brief** look we could put that blog listing in `index.rst` and tweak `index.html` to put that in the middle of the rest of template content no?

ericholscher:
http://www.writethedocs.org/conf/portland/2018/ 🎆

PR #319, opened 2017-10-04
Make sure the yaml is good

plaindocs:
Errors in the YAML are pretty hard to track down. So let's make sure the YAML is at least well formed.


ericholscher:
:heart:

PR #357, opened 2017-11-16
Include gulp and documentation needed

xpostudio4:
As requested here I include the gulp script for sass preprocessing and minifying and the documentation needed to running.


plaindocs:
Thanks! :)

PR #360, opened 2017-11-16
Give a link directly to folder with meetup data

NickVolynkin:
Current link leads to the template document which only compiles meetup information. Replaced it with a link to folder with all `meetup.yml` files.


plaindocs:
Makes sense, thanks! :+1:

PR #363, opened 2017-11-18
Lots of website fixes

ericholscher:


thatdocslady:
Wow! Awesome, looks much better, thank you @ericholscher !!!
I collected a bunch of little nits/suggestions with screenshots of examples:

- [x] Schedule overview still shows wrong alignment of titles
screenshot from 2017-11-20 08-54-23

- [ ] "Can't wait for the conf..." line should be vertical-center aligned, same as the button on the right
screenshot from 2017-11-20 08-54-27

- [ ] The "Speak" button has white text over bright green, which makes it harder to read, maybe change the text to black?
- [ ] Now that the H2, H3, H4 titles are black and have different sizing (yay!), the H1 (white over image) looks a bit small. Maybe consider making it bigger? There's space for it...
screenshot from 2017-11-20 08-54-50

- [x] The bullet list text looks more black thank the main body text, should be the same color
- [ ] The ticket types I guess is still in raw HTML, not sure who needs to fix this or how?
screenshot from 2017-11-20 08-55-29

- [ ] Bullet list paragraph alignment is weird, should be flush against the top line?
screenshot from 2017-11-20 08-55-49

- [ ] Same comment as first screenshot, the text and button both should be vertical-center aligned 
- [x] The "Portland" text here is still bold, even though the top instance is fixed (yay!)
screenshot from 2017-11-20 08-57-52

ericholscher:
Fixed the ones I could, merging now. We should send the rest over to leo.

PR #364, opened 2017-11-18
Fix blog_baseurl

maiki:
The [`blog_baseurl` doc](http://ablog.readthedocs.io/manual/ablog-configuration-options/#confval-blog_baseurl) says:

> Base URL for the website, required for generating feeds.

The extra `blog/` in the current config is adding extra path segments to the URLs in the feed, and as far as I can tell that is the only place it is affecting.

I haven't generated the site locally, so someone else please test. Until this is fixed, the URLs for items in the feeds are broken.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -24,7 +24,7 @@ extensions = [
     'ablog',
     'sphinxcontrib.datatemplates',
 ]
-blog_baseurl = 'http://www.writethedocs.org/blog/'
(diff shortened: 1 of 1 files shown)


maiki:
Ah, I forgot a preview site is generated. As far as I can tell, this only affects the feed URLs, and this change fixes it.

plaindocs:
Thanks!

Issue #368, opened 2017-11-23
Content side post launch fixes


plaindocs:
- [x] Proofread content in all newly-navigatable pages (now that the menu includes them)
- [x] Tweak navigation links (some menu items link to the wrong pages)
- [x] Link to sponsorship prospectus is broken
- [x] Tickets page fixes (also applies to Speakers page, TBD)
- [x] Fix CSS of logo hyperlink from blue > black
- [ ] heading of sponsorship prospectus is too long so it runs down to the edge of the image: 
![screenshot from 2017-12-13 14-50-03](https://user-images.githubusercontent.com/6662054/33942147-470a0cd2-e015-11e7-92f0-e04960469119.png)

thatdocslady:
Added new items to description to fix/verify. @ericholscher do you mind taking a look at these? Thx!

ericholscher:
What is the proposed fix for the sponsorship page?

thatdocslady:
I'd say either enforce a character limit on titles (not so practical) or shift the titles up so that there is more of a margin between the bottom of the text and the bottom of the image. The latter will of course affect all the other titles, but it shouldn't be a problem. I don't know exactly how many pixels we'd need to move the titles, might need to experiment a bit.

plaindocs:
I think you're talking apples and oranges here.

I'm not sure there is a fix for the sponsorship page.

ericholscher:
I also think there isn't a real issue with it. If we have an issue, perhaps we should just change the page title to something shorter?

plaindocs:
We could presumably lint the title to warn if we have too many chars :-D

plaindocs:
Are we actually using the PDF version of this page? If so something like 

```
.. only:: html

   Sponsorship - Portland 2018
   ###########################

.. only:: pdf

   Write the Docs Portland 2018 Sponsorship
   ########################################
```

otherwise just change the title? 

@ericholscher @thatdocslady

ericholscher:
Doubt it. I just printed the HTML page as a PDF when I needed a PDF :)

PR #369, opened 2017-11-27
Tickets as RST

plaindocs:
Fix for #368 

So someone needs to fix up the CSS on this PR to make it look how it should look. @thatdocslady @ericholscher ?

Two gotchas:

- [ ] Can't see an easy way of floating the price right like in the current page. Maybe

    ```rst
    .. class:: price
        450
    ```

    and some CSS wizardry to make it H2 ish ?

- [ ] Text block after sections is problematic. Hence the title hack.


thatdocslady:
I'm wondering if we're overkilling this somehow, maybe it's better to just do a clean RST text list (or use a deflist directive if it's a thing). Are we trying to make this specific section more modular/reusable?

plaindocs:
Not overthinking. I can remove classes if people prefer, and leave it formatted like the rest of the pages - but if we want to maintain the same sort of formatting as it has already we need to classify (see what I did there) the RST.

ericholscher:
I like having the little design flourishes -- I think it's what makes it distinct from our 2016 site, which was literally just a basic skin around a bunch of bare RST.

ericholscher:
eg. http://www.writethedocs.org/conf/na/2016/#tickets

plaindocs:
Thanks @xpostudio4

PR #383, opened 2017-12-01
Generified lightning talks, unconference, and tweaked generic Writing Day copy

thatdocslady:


docs/include/conf/events/writing-day.rst
```diff
@@ -1,14 +1,7 @@
-Writing Day
------------
```
ericholscher, line 2: This might break previous years, if we didn't include the headers in the main rst file.

ericholscher (commented):
PROGRESS!

PR #384, opened 2017-12-02
Updating the CFP docs in the Org Guide

kelly-o:
At long last, I've taken a pass at updating our CFP docs in the org guide. Have a look-see, and let me know if I missed/misstated anything.


ericholscher:
<3

PR #393, opened 2017-12-28
Whole lot of video updates to archives

ericholscher:


plaindocs:
http://127.0.0.1:8888/videos/eu/2015/ seems inconsistent. 
http://127.0.0.1:8888/videos/eu/2017/ is very incomplete

AU needs data file.

I'll take a bash at some of this when we get other stuff out of the way.

plaindocs:
OK, I've had another look at this, and I'm gonna push it out the door quicksharp.

I need to add talk video data to 

- [x] EU 2017
- [x] AU 2017

I need to fix up talk data from 

- [x] EU 2015
- [x] EU 2016
- [x] NA 2015
- [x] NA 2016

- [x] Looks like thumbnails are kinda proke, proably cos of https://stackoverflow.com/questions/18029550/fetch-youtube-highest-thumbnail-resolution

- [ ] At the mo we generate pages for confs that haven't happened yet

Detail pages!!!

- [x] Tons of the early ones. Looks like it might inconsistent/incomplete data in the yaml

Nice to haves but can be done after merging this in:

- [ ] Check stuff still works with Portland / Prague dirs instead of na / eu
- [x] ~Fetch talks on speaker id or something instead of numerical id.~ Nonsense, cos this is generated at iteration.

plaindocs:
Hey @ericholscher, any idea what kind of effort it would take to alter this behaviour 

https://stackoverflow.com/questions/32771287/download-html-as-separate-pages/32788289#32788289

Ie, be able to produce multiple html pages from a single rst file, say splitting into H2 or something.

plaindocs:
So the other approach would be to just generate the entire talk file, instead of leaving the datatemplate in there. It would mostly be an ugly bunch of html, to cope with the variables. 

Otherwise, it seems a *LOT* of headache is in making templates do stuff beyond `title` `content`. Maybe that is something I should look at. 

Another thought I had is we should be putting some kind of `generated from bla bla datafile, do not edit` heading in these RST files

plaindocs:
Just going off the EU 2015 YAML here, we're missing the following keys on each talks:

```
+  year: 2015
+  event: Write the Docs EU 2015
+  path: conf/eu/2015/videos/tested-and-correct-how-to-make-sure-your-documentation-keeps-working-adangoor
+  series: Write the Docs EU
+  series_slug: eu
+  slug: tested-and-correct-how-to-make-sure-your-documentation-keeps-working-adangoor
```

plaindocs:
The python script in e8de7c0 goes most of the way towards fixing up the data. (currently hardcoded to EU 2015) but adds a ton of dupe `''`, presumably from `safe_yaml`

plaindocs:
And seems to introduce some *odd* line breaks. Could be both are a round tripping issue? https://stackoverflow.com/a/36760452

Suggests using `ruamel` might improve matters?

----------------

Confirmed that `from ruamel import yaml` fixes both issues. Although we end up with quoted not literal abstracts.

docs/_ext/videos.py
```diff
@@ -24,98 +24,53 @@ def generate_video_content(session, year, region, session_idx):
...
-            print('Generated {}.rst'.format(year_path + '.rst'))
-            fp.write(generate_videolisting_by_year(year))
         for region, data in regions.items():
-            if region not in sessions_by_series:
-                sessions_by_series[region] = []
-            index_path = os.path.join('conf', region, unicode(year), 'videos')
-            shutil.rmtree(index_path, ignore_errors=True)
-            os.makedirs(index_path)
+            index_path = os.path.join('videos', region, unicode(year))
+            if not os.path.exists(index_path):
```
plaindocs, line 61: Add a flag in conf config to prevent publishing of video pages by default.

ericholscher: 👍

plaindocs:
Immortal words:

> making shit and breaking shit

PR #398, opened 2018-01-08
Add community ticket

plaindocs:


docs/conf/portland/2018/tickets.rst
```diff
@@ -4,7 +4,7 @@
 Tickets
 =======
 
-**Tickets on sale now!** Follow the links below for each ticket type. 
+**Tickets on sale now!**
```
ericholscher, line 7: I'd put a button or call out of some kind here, so people who want can just click through.

ericholscher (approved):

docs/conf/portland/2018/tickets.rst
ericholscher: I believe we should already have some defined in the HTML/CSS somewhere. Can probably just steal it from the index, unless the CSS styles are super targeted.

plaindocs: cta

Looks kinda meh if you just throw it straight in

plaindocs: I guess I could live with it floated right and vertical spacing sorted. 

Thoughts?

plaindocs:
There is also this https://ti.to/docs/widget

ericholscher:
Yea, that styling is not great :) I say just ship it, probably not a big deal as we'll sell out anyway :D

ericholscher:
We should also confirm the sponsorship address. I think it's `sponsorship@`, but I _believe_ both should work. Probably just just have it be `sponsor@`, though people might still send to `sponsors@`

PR #401, opened 2018-01-09
Lots of cleanup around how we're doing index & TOC generation

ericholscher:
This makes the front page a lot more useful,
and actually links to the content people go to!


ericholscher:
Check out the preview on netfify to feel it out.

PR #406, opened 2018-01-18
Please no warnings

plaindocs:


docs/conf/portland/2018/index.rst
```diff
@@ -1,6 +1,7 @@
 :template: 2018/na/index.html
 
-.. title:: Home - Write the Docs Portland 2018
+Home - Write the Docs Portland 2018
```
ericholscher: Did this not blow up on rendering? There's a reason I didn't use this solution, but we _could_ easily make them all as orphans.

ericholscher:
+1 on just marking all of them as orphan's for now. Not sure why that didn't occur to me :sigh:

docs/conf/portland/2018/index.rst
plaindocs: As long as we don't use the title in the body of the index page template, should be fine right?

plaindocs: I took a quick glance and it seemed ok

plaindocs:
Right, *instead* of adding them to the toc.

PR #409, opened 2018-01-30
CFP is over!

plaindocs:
Speak no more!


ericholscher:
We should probably replace these CTA's with something more useful. I guess it can be "see the schedule" once its open :)

PR #418, opened 2018-02-08
update README in regards to Python version

eirinikos:
I added a blurb regarding setting `virtualenv --python` (thanks @plaindocs) and made a few tweaks.

I think that it'd also be useful to add updates regarding commands like `make livehtml` and `make linkcheck`, but I've got some questions about those.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 This is the code that powers [www.writethedocs.org](http://www.writethedocs.org). It contains information
 about the Write the Docs group, as well as information about writing documentation.
 
-The repo is still in its early stages; feel free to contribute information that you might want to share with the community. To contribute to the Write the Docs website, famililarize yourself with the [Sphinx site generator](http://sphinx.pocoo.org/index.html).
+The repo is still in its early stages; feel free to contribute information that you might want to share with the community. To contribute to the Write the Docs website, it's helpful to familiarize yourself with the [Sphinx site generator](http://sphinx.pocoo.org/index.html), as well as [reStructuredText markup syntax](http://www.sphinx-doc.org/en/stable/rest.html).
 
 ### Prerequisites for previewing the docs locally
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,9 +15,11 @@ You'll probably need `root` privileges to install the prerequisites.
    1. download and install [pip](https://pip.pypa.io/en/stable/installing/).
    2. install `virtualenv` using your package manager.
 
+3. By default, `virtualenv` uses the Python interpreter that it was installed with. If necessary, run `virtualenv --python=/usr/bin/python2.7 <path/to/new/virtualenv>` to have `virtualenv` use the 2.7 interpreter.
```
plaindocs: "It was installed with" could possibly be confusing I think

Also "if necessary" but how do I know if it is necessary?

README.md
```diff
@@ -63,15 +65,15 @@ and instructions on how to update it are in the [`README.md`](https://github.com
 
 ### Updating CSS for the 2018 Theme
 
-The website for 2018 uses sass to compile all the assets it has, in
-order to modify the theme first you must install the dependencies of
-gulp, you can do that by running in the main directory:
+The website for 2018 uses SASS to compile all the assets it has. In
```
plaindocs: How about "To" instead of "In order to" ? (I know it was already there)

plaindocs (commented):
Great stuff! I've mentioned a couple of picky nits.

README.md
eirinikos: Cool!

README.md
```diff
@@ -63,15 +65,14 @@ and instructions on how to update it are in the [`README.md`](https://github.com
 
 ### Updating CSS for the 2018 Theme
 
-The website for 2018 uses sass to compile all the assets it has, in
-order to modify the theme first you must install the dependencies of
-gulp, you can do that by running in the main directory:
+The website for 2018 uses SASS to compile all the assets it has. To modify the theme, you must first install the dependencies of
+`gulp`. In the main directory, run:
```
eirinikos, line 69: Looks like I modified the line formatting here? Does this need to be redone?

eirinikos (commented):
Thanks! I pushed a new commit, let me know what you think. Also, I went back and re-read the README as-is - the existing info on `make livehtml` and **Previewing the docs locally** seems totally fine after all. ☺️

README.md
plaindocs: In terms of output, those two are the same.

There is some 'discussion' going on as to whether we should fully adopt a single sentence per line (probably with a soft wrap set in your editor), a hard character limit or some other arbitrary rule :-p.

plaindocs:
🎉 !

PR #422, opened 2018-02-12
update Portland 2018 schedule (remove backlinks)

eirinikos:
This change removes the green hyperlink formatting for each day's mini TOC (but preserves the permalink symbol that shows up when you mouse to the right of each heading and sub-heading).

Yay 🍪


plaindocs:
Thanks for noticing this! <3

PR #423, opened 2018-02-12
Announce speakers prep

plaindocs:
@ericholscher Grorg is cleaned up, and first batch of pics from the form are here.


plaindocs:
I've added speakers to the front page, doesn't look great yet.

ericholscher:
Let's just leave the speakers off the front page for now.

ericholscher:
Looks like the images need to be resized, but I don't want to deal with it right now. We just need to get the talks announced.

ericholscher:
Could also just kill the images, and use the names.

ericholscher:
Tried a few things. Pushed up something workable-ish with no speakers on the front page :)

plaindocs:
Lol at tooing and froing on the index page. This stuff should not really be a pain.

Anyhow, I'm happy for this to go out as is, I'm creating a separate issue for minor tweaks.

PR #430, opened 2018-02-16
Update 2018.portland.speakers.yaml

tedhudek:
Consolidated a few bullets.  Note that is now a mismatch between the six points referenced in the title and the eight in some of the metadata. Hopefully that won't cause any issues.  Thanks!


plaindocs:
Cheers! That should not cause any problems, but we could also just fix the metadata, as I'm pretty sure it is used both to construct the link and the anchor.

tedhudek:
Thanks @plaindocs looks great

PR #433, opened 2018-02-21
THE MERGE take 2

plaindocs:
This is branched from the previous PR, which might have been a mistake. 

* Speakers are up (flag toggle works, especially if you put the right variable in)
* Can't quite figure out the variables in the `path to` stuff, @ericholscher wanna poke those? (or just leave hard coded) ?

And then I'm not quite sure why GH is confused if I wanna put these four commits onto master. 

MAYBE it is cos I deleted the branch after merging? ??

### DIFF
--- a/docs/_data/config-portland-2018.yaml
+++ b/docs/_data/config-portland-2018.yaml
@@ -40,8 +40,8 @@ cfp:
   notification: "end of January"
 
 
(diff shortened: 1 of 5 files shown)


ericholscher:
I went ahead and reverted the revert, and then merge this in locally. 👍

PR #434, opened 2018-02-22
Update tours and welcome wagon intro time

canncrochet:
Change the tours to happen during Welcome Wagon intro and Monday morning only. Change Welcome Wagon intro to 5:00 pm to coincide with the start of the opening reception.


ericholscher:
🎆

Issue #451, opened 2018-03-14
Upgrade python code to 3.5?


plaindocs:
http://python-future.org/automatic_conversion.html

plaindocs:
I'm gonna call this done!

PR #452, opened 2018-03-14
Switch to Python 3.5

plaindocs:
Address  #451

So this fixes the Netlify problem which is good to know.

I've tested the combo of Python 3.5 *and* Video archive over in 
https://github.com/writethedocs/www/tree/sw-python-35-and-videos and #453 

~I've used Pipfile for dependencies, rather than fiddling with multiple venvs, should disentangle that before merging. Also seems ok. Not sure about merging both of these in one go.~ :trollface:


plaindocs:
I've diffed the html output and it looks mostly fine. 
Some unicode stuff that still displays ok as html.

ericholscher:
Looks good to me! I'm a fan of moving more of my life to py3. We should also update our install docs though.

plaindocs:
Yeah, I'll update the README and merge.

Thoughts on Pipfile vs requirements.txt?

ericholscher:
People seem to like pipfile, I haven't really used it. I'm fine switching to it, as long as we document how to use it :)

ericholscher:
(Though I don't think RTD supports it yet)

plaindocs:
Right, in that case `requirements.txt` it is.

plaindocs:
:shipit: ?

ericholscher:
👍

PR #453, opened 2018-03-14
Video archive, updated to python 3.6

plaindocs:
Just testing.


plaindocs:
Urgh, I was hoping to avoid that conflict.

plaindocs:
We should probably just merge this as well, cos `Files changed 286`.

What about generating and caching all the talk files somehow, without committing them to the repo? Sphinx's 1 input : 1 output file is a *real* limitation here. With another static site generator you could just define the path and generate directly from the YAML.

ericholscher:
We could output the files outside of Sphinx if we wanted. Getting them in the theme & TOC would be tricky -- how do other platforms handle that?

plaindocs:
All of these have trade offs. 

Without the 1input : 1output limitation, you just generate everything under `/videos/` drectly from the YAML, with different templates. With Sphinx, instead of adding 200 video `*.rst` files to the repository, generate them at build time, and process with sphinx. Of course, once we've added them, they will rarely or never change, so same same.

And I don't know how caching would work, exactly, especially across RTD/Netlify/Anything else, but if you generated the output with sphinx, and included it as static files you would not have to generate it at every build, regardless of whether you were building from rst or directly from yaml.

plaindocs:
Just spitballing really. Either way, I'd *love* to merge this video archive.

ericholscher:
+1 on merging it. I'm at the end of my day, but feel free to merge it whenever you have time to deal w/ any fallout!

ericholscher (approved):

ericholscher:
We should also make sure travis is running w/ 3 here

plaindocs:
Good catch!

ericholscher:
K. tired of waiting for videos to build locally, merging!

PR #458, opened 2018-03-20
Rework NA sponsorship templates to use a single source

ericholscher:
This keeps the sponsorship data in a minimal set of HTML files with just the sponsorship fragments.
Then anything going around the sponsorship data is kept in the place of inclusion.

This also adds sponsors to the footer of every page (except the sponsorship page),
so they get more visibility.


ericholscher:
Gonna merge this @plaindocs because we just announced the sponsors in this email, but take a peek later.

plaindocs:
I'm somewhat confused by the logo sizing and organising here. I'd like us to be clear about sizings and columns, also re the interfacing of our stated sponsorship levels and the gold/silver from the templating. 

Also, what does the `macro` do?

ericholscher:
> I'm somewhat confused by the logo sizing and organising here. I'd like us to be clear about sizings and columns, also re the interfacing of our stated sponsorship levels and the gold/silver from the templating.

The image resizing is just so we don't ship massive images down to the browsers and have standardish sizes, nothing about sponsorship levels really. 

The problem is that there's only two levels of sponsorship in the CSS from what I can tell, so I wanted a way to separate the medium & small sponsorships. 

 > Also, what does the macro do?

The macro is so we can indent the output.

plaindocs:
Yeah, understood about the resizing. I was more asking about the next point

Let's chat about the sponsorship levels separately, on slack or 1:1

Macro: gotcha. :+1:

PR #461, opened 2018-03-21
Unbreak the display & updating of news on the front page.

ericholscher:


plaindocs:
Amazing! Thanks for this!

Do we want to attempt to use the template news look for this, or do we prefer to just have the automated list?

ericholscher:
I think an automated list is the only way it stays up to date. Filling out that janky HTML is not a good solution.

plaindocs:
I'm 100% in agreement with that. 

What I mean is, do we try to kick the way ablog displays into how it looked before, by giving it a few CSS classes and what not? Or do we just accept the look as is.

Issue #465, opened 2018-03-22
Fix broken links


plaindocs:
Just wanted to let you know that there are a few dead links on your style guide [http://www.writethedocs.org/guide/writing/style-guides/] page.

- [x] Pebble Road [https://www.pebbleroad.com/articles/Creating-Maintaining-a-Web-Style-Guide]
- [x] Pressly [http://blog.pressly.com/2013/09/24/5-style-guides-craft-content/]
- [x] IBM [https://www.redbooks.ibm.com/Redbooks.nsf/ibmpressisbn/9780132101301?Open]

plaindocs:
@eirinikos another nice easy one :-D

eirinikos:
thanks @plaindocs !! ❤️

PR #467, opened 2018-03-24
update IBM style guide link

eirinikos:
(addresses #465 )

In regards to the IBM link, I also found a link to a [PDF](http://ptgmedia.pearsoncmg.com/images/9780132101301/samplepages/0132101300.pdf).

In regards to the Pebble Road and Pressly links, I did some googling ...
- [This](https://www.pebbleroad.com/articles) seems to be the extent of all Pebble Road articles that are currently available to the public.
- [Pressly](https://www.pressly.com/) appears to have been acquired by another entity. Not sure what this means for its previously existing blog posts.

What are your thoughts? Should we just remove those latter two links? And should we notify Ivan Walsh? 🤔


plaindocs:
Awesome stuff!

If you can't easily find the link, just delete the links. And no need to notify Ivan I don't think.

PR #469, opened 2018-03-25
Update team.rst

diggerlilly:
Add Alicia to Explore Portland role. 
(Hmm, could I have created a single PR containing both this change and my alicia.rst include? Sorry for extra work if so.)


plaindocs:
Yep, a single PR would have made more sense. But this works fine too!

PR #449, opened 2018-03-13
update vale config with EU rule

eirinikos:

PR #472, opened 2018-03-27
Update vale checks take2

plaindocs:
Replaces #449


plaindocs:
~So I think tox is doing something really weird, cos I'm getting different results for running vale locally, running vale in tox, locally and the tox running on travis ci...~

~Which is *really* unhelpful for what we're doing here.~

Gah, tox builds to no standard output dir in .tox

ericholscher:
We should look at merging this. It looks useful :)

ericholscher:
@plaindocs this still worth poking at? ^^

plaindocs:
Yeah, it sadly isn't really now we're making almost everything use the variables. What might be more interesting is to ignore the *source* and lint the *output* html.

But to do that we'd have to configure sphinx to pass the vale "ignore this bit" comments though to the output, and it all gets pretty fiddly.

PR #474, opened 2018-04-02
Make the templates for 2018 generic.

ericholscher:
This was done when adding the Cincy conf to the site. Now we don’t have
city-specific templating, but just a base set of templates for each year,
except for the menus (for now).

This allows us to generate all the interesting bits & pieces from the YAML
file, without having to customize each set of templates.


plaindocs:
Ooh nice!!

This looks like we should be able to delete 

```
eu/generic.html
eu/index.html
na/generic.html
na/index.html
base_na.html
base_eu.html
```

and refactor till things aren't broken, right? 

Then we need to fixup the speaker template and move it out of NA

plaindocs (approved):
Cincinnatti looks great.

Do we need a newsletter segment? 
And we should probably automate the list of confs as well at some point. :-p

plaindocs:
I also like the data format for the three days  -although we'll need to figure some more standardization for that.

ericholscher:
Yea we should be able to delete all the customized templates and use flags, I just didn't get around to the menus yet.

plaindocs:
I'm working on this fixups here btw.

plaindocs:
Urgh, waat. Those [conflicting files] haven't been changed for ages.

plaindocs:
They are just deletions anyway I think.

ericholscher:
@plaindocs think we should ship this? Cincy isn't linked up yet, but I want to get this out before the templates diverge more.

ericholscher:
heh, not quite ready yet: 

screenshot 2018-04-10 12 42 03

plaindocs:
HA! I had to fix a couple of the conditions on the templates already, because non existing key seems to test as true. :-p

I'll take a look.

:+1: to merging sooner rather than later.

ericholscher:
Yea, not sure where those dates are coming from. Perhaps we're not properly overwriting the context variables or something.

plaindocs:
OK, I *think* we can merge this - wanna take it for a quick spin beforehand?

plaindocs:
:+1:

PR #476, opened 2018-04-05
Fix title of Ted Hudek's talk

ghost:
Tuesday schedule lists Ted Hudek's talk as
"Eight Essential Tips for the Enlightened Tech Writer":
http://www.writethedocs.org/conf/portland/2018/schedule/#id4

But the Speakers page lists six tips:
http://www.writethedocs.org/conf/portland/2018/speakers/#speaker-portland-2018-ted-hudek


plaindocs:
Thanks Ted! We'd already fixed it in the original speaker list, but reintroduced the revision in the schedule. :-(

PR #479, opened 2018-04-06
Test out stage2 yaml

plaindocs:
A slight improvement?


ericholscher:
Hmm. I tried to merge this with the existing stuff locally, and I'm not sure if it really works. I think we don't really need a fully broken about schedule for lola's. I think the existing listing of it works fine, and takes up much less room.

plaindocs:
Yeah, it was always more of a lets try this and see. I'm happy to close this, or migrate some of the heading squishing manually.

PR #495, opened 2018-04-23
Try to fix dates

bethaitman:
I'm not sure if this is the best way to fix the thing with the dates, but I've given it a go.

### DIFF
--- a/docs/_templates/2018/index.html
+++ b/docs/_templates/2018/index.html
@@ -126,10 +126,10 @@ Home - Write the Docs {{ name }} {{ year }}
         <p>{{ date.day_two.summary }} </p>
       </div>
       <div class="schedule-item uk-width-1-3@s">
-        <img style="height: 150px;" src="{{ pathto('_static/2018/assets/icons/' + date.day_three.icon + '-green.svg', 1) }}">
-        <h3>{{ date.day_three.event }} </h3>
-        <em>{{ date.day_three.date }} </em>
-        <p>{{ date.day_three.summary }} </p>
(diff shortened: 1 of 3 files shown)


docs/_templates/2018/index.html
```diff
@@ -126,10 +126,10 @@ <h3>{{ date.day_two.event }} </h3>
...
       </div>
       <div class="schedule-item uk-width-1-3@s">
-        <img style="height: 150px;" src="{{ pathto('_static/2018/assets/icons/' + date.day_three.icon + '-green.svg', 1) }}">
-        <h3>{{ date.day_three.event }} </h3>
-        <em>{{ date.day_three.date }} </em>
-        <p>{{ date.day_three.summary }} </p>
+        <img style="height: 150px;" src="{{ pathto('_static/2018/assets/icons/' + date.main_conference.icon + '-green.svg', 1) }}">
+        <h3>{{ date.main_conference.event }} </h3>
+        <em>{{ date.main_conference.date }} </em>
+        <p>{{ date.main_conference.summary }} </p>
```
ericholscher, line 132: This will break on all our other confs, unless we update all the YAML.

ericholscher (commented):
Thanks for the PR beth! I think what we've been doing here is the wrong approach though. We should instead have:

* A days array that holds the days of the conference
* Then iterate over the days for the schedule, and output them without a "day one, day two" concept
* Add a "schedule overview" for the index, which is likely separate from the full listings of days, which would result in the schedule. This could be the existing "day_one/day_two/day_three" concept, but named a bit differently.

Issue #501, opened 2018-04-26
Clarify whether non-conference attendees are allowed to come to the Sunday Evening reception


mjang:
ref http://www.writethedocs.org/conf/portland/2018/schedule/#sunday-may-6, which is not clear on the subject, and a recent Q in #wtd-conferences in the Slack channel

annegentle:
I'll have a go at it 😀

PR #502, opened 2018-04-26
Updated with a note about additional guests

annegentle:
This edit is for the Sunday reception only - see what you think.

Addresses Issue #501


plaindocs:
Many thanks @annegentle :-)

PR #507, opened 2018-04-26
Rename au -> australia for 2018

ericholscher:


ericholscher:
@plaindocs renamed the `au` dir for 2018, and incorporated your PR.

docs/_data/config-australia-2018.yaml
```diff
@@ -1,6 +1,6 @@
 name: Australia
 
-shortcode: au # prague australia?
+shortcode: australia # prague australia?
```
plaindocs, line 3: The only reason that we have shortcode and template code is that there are `na` and `eu` in a few places in the templates and datafiles. We should probably kill those once and for all and clean up at some point

plaindocs (approved):
Sounds good to me.

docs/_data/config-australia-2018.yaml
ericholscher: Aye. The shortcode is used a lot of places, but it's basically just `name|lower`.

plaindocs: Right. It is now. But not when we had Portland, `na`. Which is why we have the hacky template code.

PR #510, opened 2018-04-27
Fix ticket links

shaunix:
We used a different slug on tito than shortcode, so the ticket links don't work.

I don't know if this is the best way to fix this. We can't just change shortcode, because other stuff would break.


ericholscher:
:+1:

plaindocs:
Yeah, consistency isn't always easy :/

PR #563, opened 2018-05-06
Add organizer and other meta data to meetup listing page

agjohnson:
Organizer information is not complete and doesn't necessarily seem to match up
to Meetup data. We could automate this by hitting the Meetup API during build
and grabbing all of the organizer information.

Refs #54

### DIFF
--- a/docs/_data/meetups/aus.yaml
+++ b/docs/_data/meetups/aus.yaml
@@ -1,3 +1,6 @@
 region: Oceania
 country: Australia
 website: http://www.meetup.com/Write-the-Docs-Australia/
+meetup: Write-the-Docs-Australia
+organizers:
+  - name: Swapnil Ogale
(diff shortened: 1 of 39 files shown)


plaindocs:
Nice!

plaindocs:
I'm gonna merge this in, and maybe @mjang wants to shout out in #meetup-organisers in case people wanna update or add their info.

@agjohnson Can you even get user data from the API? I thought maybe not. Or maybe you can just not the email.

agjohnson:
It looks possible, but seems to require iterating over *all* members of the group.

https://secure.meetup.com/meetup_api/console/?path=/:urlname/members

Python query would be something like:

```python
organizers = [
    profile for profile in response
    if (
        profile['group_profile']['status'] == 'active' and
        profile['group_profile']['role'] == 'coorganizer'
    )
]
```

mjang:
Thanks -- it's great to include organizer names! (But there are some updates that I'll try to PR this evening or maybe this weekend -- then I'll announce in #meetup-organizers)

PR #582, opened 2018-05-25
Update name and organisers

StarfallProjects:
Updated to reflect new name, geographical area and additional organiser


plaindocs:
That is a proper cheeky meetup.com URL!

PR #584, opened 2018-05-31
Update for jobs channel rename job-posts-only

nschonni:
I'm not sure if this is still an "auto-join" room, but it was renamed recently


plaindocs:
:+1: thanks!

PR #586, opened 2018-05-31
Bump sphinx and ablog

plaindocs:


plaindocs:
@ericholscher I've bumped some version numbers to see if it fixes the build problem. Any objctions to merging?

ericholscher:
Nope, looks good.

ericholscher:
@plaindocs is there a specific issue that this is fixing? Not sure why it'd randomly start breaking, unless we're doing something new :)

plaindocs:
I can't see a cause. There is an error in the atom.xml generation. But I'm at a bit of a loss as to why. Works OK locally, and on netlify. Take a look at the RTD build log and see if you have a better suggestion?

ericholscher:
Looks like this is the issue: https://github.com/abakan/ablog/issues/74#issuecomment-369167516

Not sure why it randomly started happening now though :/

ericholscher:
Oh. I bet the ablog fork we were using got updated. Didn't realize we were pinning it to github :/

plaindocs:
Maybe, don't think we were pinning to a commit.

ericholscher:
Of course bumping sphinx caused some other issues -- fixing :)

ericholscher:
The website should now be updated properly at least 👍

PR #588, opened 2018-06-01
Slack/issue587

mjang:


mjang:
The description for why we do the Slack Free plan is inspired by Monique S -- and I have her permission.

mjang:
IMO, we have some recent instances of Spam, and I think it'd be helpful to have an explicit limitation on such on our Slack page.

docs/slack.rst
```diff
@@ -11,6 +11,11 @@ Below we have a bit more information about the useful Slack channels that we hav
 This should start you off on the right foot,
 and get you connected with folks who have similar interests.
 
+Our network uses the Slack "Free" plan. Under this plan, Slack limits access to the latest
+10k messages. For us, this is a feature, as many people seem to feel a freedom in knowing
+that their words will disappear shortly. And many have said that if the conversations were
+archived/published, they’d quit participating.
+
```
Bradamant3: I might rephrase this a bit, because cost is also a factor, and I think folks are pretty blissfully unaware of just how expensive paid Slack can be. (also please remove quotes around "free" ;) ) Suggestion:

"We don't pay for our Slack, which means that the community has access to only the most recent 10,000 messages. Not only is paid Slack prohibitively expensive for a community like ours, but many people seem to feel a freedom in knowing that their words will disappear shortly. Folks have also said that if the conversations were archived/published, they’d quit participating."

(we could expand, but that would probably get into too many more words, and this is mostly for reference anyway, right? Most people won't read it, and we'll undoubtedly continue to get questions about it, which we'll continue to field. In the same spirit as the conversation that inspired this addition :D )

docs/slack.rst
```diff
@@ -75,6 +80,7 @@ please avoid these behaviors:
 - **Let Me Google That For You (LMGTFY)** - If people ask questions that are simple to answer on Google, still answer them nicely.
 - **Question Policing** - Some people might ask incomplete questions that are hard to answer. Ask follow up questions and act in good faith to help them solve their issues.
 - **Technology Competitions** - We don't care if your tool is faster or better in some metric. We care about creating great documentation for our users.
+- **Spam** - Respect our space. Don't post spam or otherwise vandalize our community.
```
Bradamant3, line 84: I don't think this item addresses the post/s that troubled you recently, Mike. IOW, I doubt that the poster would recognize their post as spam. There's a slippery slope with contributors falling into promotional posts, but up to now it's been OK bc said posters have also been real participants in the community. The recent incident wasn't like this, and I think we want to discourage such things, but I'm not sure this will help.

Bradamant3 (commented):
Added a couple of comments. More to discuss?

docs/slack.rst
mjang: How would you address such promotional posts? 

The original post that I was concerned about included the following phrase: "we’re constantly posting across a variety of platforms " I think that's a pretty explicit acknowledgement of spam.

docs/slack.rst
mjang: wfm, will update as such.

mjang:
Merging, based on @Bradamant3 's feedback re our Slack plan. I think the generic bit about spam is OK too -- but if someone disagrees, I'll be happy to open a new PR to change or revert.

PR #589, opened 2018-06-02
Prague welcome wagon cleanup

ddbeck:
I did a read through of the Welcome Wagon info and updated a couple of things:

* Added @bethaitman's name and mine to the relevant locations
* Dropped a paragraph about contacting accounts I don't control
* Fixed a couple of links to point to their 2018 equivalents


ericholscher:
Looks great, thanks :)

PR #592, opened 2018-06-06
Fixes typos and links; removes some irrelevant info.

sarahmaddox:
I've also shortened a couple of sentences, changed some upper case to lower case, and changed passive voice to active.


plaindocs:
Looks great, thanks!

PR #608, opened 2018-06-25
Fix speaker export to be py2

ericholscher:
Heroku is py2 currently, and py3 is broken.
Hopefully we'll never have to use this again :)


plaindocs:
LGTM

PR #615, opened 2018-07-05
Create ind-bangalore

Rajakavitha1:
Hi, Here's a meetup group that we have in India. Could you please add this to the list.


mjang (approved):
Thanks!

PR #623, opened 2018-07-16
Update Jessica Garson's talk for cin2018

shaunix:


shaunix:
Sorry, looks like Hillary already handled this in #622. Let's use hers, as it has another update.

PR #627, opened 2018-07-19
Revert "fix amw slug to match slug in speaker list"

hillaryfraley:
Reverts writethedocs/www#626


hillaryfraley:
I merged Adam's PR, but something didn't seem right on the site after I did, so there might be something else that has to be fixed along with this change.

edit: never mind, I think it's all OK. The link in the schedule to the talk description didn't seem to be working, but now it's fine.

PR #653, opened 2018-08-15
Revert "Au conf so"

ericholscher:
Reverts writethedocs/www#646


plaindocs:
https://github.com/writethedocs/www/pull/646/commits/7b5ebce4db3e5f213b219368d928a8d482b8b9b1 
looks like most of the damage.

PR #679, opened 2018-09-14
Prague Post Conf Updates + Videos

plaindocs:


ericholscher (approved):
Looks good. I do want to define the conference states a bit more explicitly, but this is good until we do that 👍

Issue #687, opened 2018-10-03
Include interview Qs on WtD Website


mjang:
Notes from Kristy James, from a WtD '18 EU unconf session

Discussed in the WtD Slack, #career-advice 2 Oct. I've copied the notes below. 

I think this would make a great addition to the WtD website, 

perhaps under "[Learning Resources](http://www.writethedocs.org/about/learning-resources/#)" - or with the [Job Board](http://www.writethedocs.org/job-board/)

# List of questions to use in job interviews #

## Self-starter ##
How would you approach a [task]?

Can you tell me about a time you initiated a project?

What would you do if you want to perform a project but don't get support from your teammates?

## Team player ##
What do you imagine your ideal team being?

How would you go about teaching a new skill to your team?

## Organizational experience
If this were a new department, where would you start?
Can you tell me about a time you showed leadership changing a process/procedure?

Do you think a writer should only right, or should they edit?
How would you deal with the situation when senior staff members keep sending  you (too many) emails to edit?

What's the thing you're most proud of in your professional life?
What's the hardest lesson you've had to learn?

# Dealing with feedback #

How would you deal with a senior member of your team who is giving you substandard text but you know they can do better?

How to deal with feedback from a senior writer about your style when they are criticising it but you believe you're right?

If a developer comes to you and they have to write a document, e.g. a readme, how would you get them started?

How do you keep up your concentration when you have a monotonous task?

## Personality? ##

How would your colleagues describe you?

# Bad questions #

Very in-depth questions about the subject matter, not about the writing skills.

Who *are* you and what's your passion?

What's your five year plan?  -> What's the next thing you want to learn?

What's your favourite food?

What would be your superpower? -> Would you rather be able to fly or be invisible (see reasoning)

# What questions do tech writers need to ask companies when interviewing #

What do you think a tech writer does?

How much influence do you see the writer having within the team? How would you help achieve this if it doesn't happen organically?  Are they allowed to give input in other places?

Would you support writer suggesting process changes/UX changes if they see the need?

How would you evaluate the success of a TW?

Could you descibe your company culture - from the perspective of a new graduate hire and a new joiner with 15 years' professional experience?

What value do you think (improved) documentation will bring to your company?  Who will your audience be? How will you measure this value?

What training would you provide to educate a new employee about the context of the business (which depts, the product, existing systems) -> Red flags - do not know or not going to tell
What training would you provide/support for professional development?

Where do you see this role fitting between a focus on Users, Business, Technology triangle

[...]

PR #690, opened 2018-10-08
687 Add "Hiring Guide" page under Learning Resources menu item with l…

kristyj:
Closes issue #687 - adds the list of interview question suggestions brainstormed at WTD Prague 2018 to the website.
Note: Leaves room for expansion of this new section for other questions discussed on the career-advice channel.

### DIFF
--- a/docs/hiring-guide/index.rst
+++ b/docs/hiring-guide/index.rst
@@ -1,6 +1,6 @@
-============
-Hiring Guide
-============
+=================================
+Guide to Hiring and Getting Hired
+=================================
 
 Welcome! This guide aims to help those looking to hire documentarians find the perfect candidate, and to help job-seeking documentarians find the perfect position.
 Its content focuses on tricky points in the process of hiring a documentarian.
--- a/docs/hiring-guide/index.rst
+++ b/docs/hiring-guide/index.rst
@@ -1,0 +1,102 @@
+============
+Hiring Guide
+============
+
+Welcome! This guide aims to help those looking to hire documentarians find the perfect candidate, and to help job-seeking documentarians find the perfect position.
+Its content focuses on tricky points in the process of hiring a documentarian.
+
+A great place to start on your hiring or job-seeking journey is by posting to or browsing the Job Board.
+Check it out over at https://jobs.writethedocs.org/.
+
+.. _job board: https://jobs.writethedocs.org/
+
+
+.. note:: This is a living (and growing) guide that is missing some topics.
+          If you want to contribute, you can do so by
+          creating `issues <https://github.com/writethedocs/www>`_ and
+          `pull requests <https://github.com/writethedocs/www/pulls>`_ in the
+          `GitHub repository <https://github.com/writethedocs/www>`_.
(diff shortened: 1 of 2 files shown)


plaindocs:
This is great! 

I'm not a 100% sure about creating a new Guide rather than just adding a Hiring heading to the main Guide, what do you think?

When we do merge it in, I'd go as far as suggesting pinging the newsletter folks and asking for a mention.

plaindocs (approved):

ericholscher:
> I'm not a 100% sure about creating a new Guide rather than just adding a Hiring heading to the main Guide, what do you think?

Yea, I'd just include it under the main guide, until we have enough stuff to justify breaking it out.

ericholscher:
Actually, I'm having second thoughts. I do think career/hiring is a different topic than the "doc guide" that we have. I do think it maybe makes sense to break out as a different guide. 

I wonder if having it as a different section will also make it easier to contribute to, and build out as a contained resource.

ericholscher:
Also, thanks so much for this PR @kristyj! It's 💯

plaindocs:
I can see both sides of this. "Documentation Culture" that is already included in the Guide should probably live by the same logic? Kinda depends if the Guide is about writing Docs or about Doc related stuff... 

> This guide exists to provide both novice and expert writers a best practice handbook for building, structuring, and writing software documentation.

docs/hiring-guide/index.rst
```diff
@@ -0,0 +1,102 @@
+============
+Hiring Guide
```
mjang: Great contribution!

When I read "Hiring Guide", I think this advice is limited to hiring managers and HR, so I'm trying to think of a different title. (and that's my only concern with the PR)

Maybe "Guide to Hiring and Getting Hired"?

mjang: Or maybe "Hiring and Interviewing" (brainstorming -- I have an open mind on this)

kristyj: I was definitely wrestling with the question of the name too @mjang. I like your suggestion of "Guide to Hiring and Getting Hired" but perhaps it's a little long for the sidebar - how would you feel about keeping it the same in the sidebar but using the new title on the page itself?

mjang (approved):
wfm, thanks!

kristyj:
Thanks all for your reviews so far! @ericholscher I did feel that the "Documentation guide" section gives a strong first impression that it a resource only for writing advice and tools, hence wanting to split this topic into a separate section. The audience may be hiring managers who do not know much about docs and will need a visible location to access this information. 

> I wonder if having it as a different section will also make it easier to contribute to, and build out as a contained resource.

I recognise that it seems a bit bare now, but it would be great if we could encourage contributors! There seems to be a lot of activity on slack that could be added here for posterity.

ericholscher:
I think this looks great for now, and agree we can iterate on it over time. I'm gonna merge it to get the info out there, and we can work on it more over time with it live. 🎉 

Thanks for the PR/input folks!

PR #703, opened 2018-10-17
Update index.rst

firstladymb:
Fixing broken sponsor image for mdn. Referencing the 2016 version (/_static/img/2016/sponsors/mdn.png) instead since it doesn't exist in the 2015 folder.


firstladymb:
Found other links that need fixing on the page. Going to do them in one branch!

PR #709, opened 2018-10-17
Force default file encoding to UTF8 on windows

grisevg:
Force default file encoding to UTF8 on windows. Otherwise it crashes when opening non-ascii files on windows.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -7,6 +7,11 @@ import ablog
 import sys
 import os
 
+# Only for windows compatability - Forces default encoding to UTF8, which it may not be on windows
+if os.name == 'nt':
+    reload(sys)
+    sys.setdefaultencoding('UTF8')
+
 
 sys.path.append(os.getcwd())  # noqa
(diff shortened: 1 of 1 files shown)


plaindocs:
@ericholscher you know of any reason not to do this?

grisevg:
I'm really not sure if it's correct thing to do or not
This old SO post (from 2010) says it's bad https://stackoverflow.com/questions/3828723/why-should-we-not-use-sys-setdefaultencodingutf-8-in-a-py-script

ericholscher:
It sounds like this shouldn't be necessary with python3 (and the SO post says it doesn't do anything). Were you running this with python3?

I don't have access to a windows machine, but I'm happy to include this if it fixes things for windows users though.

grisevg:
Yes, I tried both Python 3.6 and Python 3.7, same thing on both.

ericholscher (approved):
Seems worth trying. :+1:

PR #712, opened 2018-10-22
Update Podcast URL to https

tomjoht:


ericholscher (approved):
Yay HTTPS! We should update our internal links as well. 🎉

PR #718, opened 2018-10-30
New SVG web icons coral and blue

thatdocslady:
BTW I'm not sure why the whole thing is inside a 2018 folder, maybe @ericholscher you can rejigg the hierarchy without imploding the whole repo? I didn't want to touch it but I created new folders for the subsequent icon years


ericholscher:
Merged this into #721 and it will go live there. 👍

PR #726, opened 2018-11-14
Update index.rst

debri174:
Fixed broken unconference links by removing the links as there's no content to link to and this conference is 3 years old.


plaindocs (approved):
:+1:

PR #727, opened 2018-11-14
Update menu-mobile.html

debri174:
Fixed Prospectus link in mobile nav for both Cincinnati and Australia


plaindocs (approved):
Awesome, thank you.

Making new conferences is still a slightly manual process, so errors from one often propagate to the next conf and live on till they are spotted and fixed.

PR #731, opened 2018-11-14
HTMLProof this!

plaindocs:

PR #732, opened 2018-11-14
Does this link check?

plaindocs:
Replaces #731 to avoid merge conflict. 

Seems a bit silly to build the HTML twice though, we should be able to share those between tox envs?


plaindocs:
Oh wow, this actually works now! I just need to winnow down the amount of failures, then we can start fixing.

plaindocs:
~Ewwww, I must have messed up that merge?~

plaindocs:
We have a a few broad categories of link failures:
  - Links to speaker headings in various places (these will need fixing in the templating / yaml not one by one) 
  - Some links to news and blog posts, not really sure why these are broke. 
  - Some path links, to `news` which should be `../news` etc,have not checked if these are errors in the template, but sounds probably. Ditto `sponsors`. Probably need to be using absolute paths in all templates? 
  - A fairly small number of external failing links

plaindocs:
And in 5b659c3 I've set the link checker job to fail without affecting the build!

ericholscher (approved):
Changes look good. Seems we just need to whittle down the links until it passes.

PR #733, opened 2018-11-22
HTML proofer is getting the paths wrong

plaindocs:
The link checking we merged in #732 isn't getting the paths right. 

Something is getting confused, perhaps `pathto` in the themes. For example:
```
<li><a href="{{ pathto('conf') }}">Past Events</a></li>
``` 
from a menu, which we serve as `/conf/` is getting link checked as `conf.html` and results in a 404. I was playing around with `html_link_suffix = '/'` in `conf.py` which doesn't fix the problem.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -39,7 +39,11 @@ commands =
 changedir = {toxinidir}/docs
 whitelist_externals = htmlproofer
                       gem
-commands =
+
+commands_pre =
     gem install html-proofer
+
+
+commands =
     sphinx-build -b html -d {envtmpdir}/doctrees . {envtmpdir}/html
-    htmlproofer {envtmpdir}/html --alt-ignore '/.*/' --file-ignore '/2013/','/2014'/,'/2015/' --url-ignore '/video/' --allow-hash-href
+    htmlproofer {envtmpdir}/html/conf/australia/2018/about.html --alt-ignore '/.*/' --file-ignore "/2013/,/2014/,/2015/,/blog/" --url-swap ".html:/" --disable_external --url-ignore '/video/' --allow-hash-href
--- a/tox.ini
+++ b/tox.ini
@@ -45,4 +45,4 @@ commands_pre =
 
 commands =
     sphinx-build -b dirhtml -d {envtmpdir}/doctrees . {envtmpdir}/html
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -39,7 +39,11 @@ commands =
...
-commands =
+
+commands_pre =
     gem install html-proofer
+
+
+commands =
     sphinx-build -b html -d {envtmpdir}/doctrees . {envtmpdir}/html
-    htmlproofer {envtmpdir}/html --alt-ignore '/.*/' --file-ignore '/2013/','/2014'/,'/2015/' --url-ignore '/video/' --allow-hash-href
+    htmlproofer {envtmpdir}/html/conf/australia/2018/about.html --alt-ignore '/.*/' --file-ignore "/2013/,/2014/,/2015/,/blog/" --url-swap ".html:/" --disable_external --url-ignore '/video/' --allow-hash-href
```
ericholscher: DO we want the full path to `conf/australia/2018/about.html` here?

plaindocs: That is just me testing on a single file, so I don't have to scroll through oodles of results.

plaindocs: Will remove once I understand how to fix the problem.

plaindocs:
@ericholscher this is now working, and after ignoring vast swathes of 2013 and 2014 is down to a very manageable number of broken links.

tox.ini
```diff
@@ -39,7 +39,10 @@ commands =
...
-commands =
+
+commands_pre =
     gem install html-proofer
-    sphinx-build -b html -d {envtmpdir}/doctrees . {envtmpdir}/html
-    htmlproofer {envtmpdir}/html --alt-ignore '/.*/' --file-ignore '/2013/','/2014'/,'/2015/' --url-ignore '/video/' --allow-hash-href
+
+commands =
+    sphinx-build -b dirhtml -d {envtmpdir}/doctrees . {envtmpdir}/html
+    htmlproofer {envtmpdir}/html --alt-ignore '/.*/' --file-ignore -file-ignore "/2013/,/2014/,/2015/," --url-ignore '/2015/,/video/' --allow-hash-href
```
ericholscher: `--file-ignore -file-ignore` feels like we're probably doing something wrong.

ericholscher (approved):
Looks good, with one small nit.

tox.ini
plaindocs: gah, thanks!

PR #738, opened 2018-11-29
Move to using Jinja includes instead of rst

ericholscher:
This allows us to use our Jinja variables in the include files

### DIFF
--- a/docs/about/events-activities.rst
+++ b/docs/about/events-activities.rst
@@ -13,7 +13,7 @@ Conferences
 
 Our conferences create a time and a place for the global community of documentarians to share information, discuss ideas, and work together to improve the art and science of documentation.
 
-.. include:: /include/conf/current.rst
+{% include "conf/current.rst" %}
 
 Meetups
 ~~~~~~~
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -53,7 +53,7 @@ blog_locations = {
 blog_default_location = 'PDX'
 fontawesome_link_cdn = 'https://netdna.bootstrapcdn.com/font-awesome/4.0.3/css/font-awesome.min.css'
 
-templates_path = ['_templates']
+templates_path = ['_templates', 'include']
 templates_path.append(ablog.get_html_templates_path())
 
 source_parsers = {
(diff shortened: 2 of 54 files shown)


plaindocs:
Looks great. Next steps, move data like sponsors, dates to the data file, and consolidate all of the templates.

Then templatize all of that data in the includes files for CFP, Writing Day, etc.

plaindocs:
And we can probably just IF pre 2019 and IF 2019 so as not to back port?

ericholscher:
Yea we should be able to not backport. I think we also need to think more deeply about our YAML structure before we do a full refactor here too. I want to figure out a better structure for the dates/schedule data.

plaindocs:
Agree. But I don't think the two things are tightly connected.

ericholscher:
Well, the way we structure the schedule & front page does rely heavily on that data, but agreed the other pages should be safe.

plaindocs:
Dates being easily reusable *and* human readable / editable is a tough combo.

plaindocs:
So we tried something like: 

```
date:
  year: 2019
  month: May
  day_1
    dom: 19
    main_event: {{ social }}
  day_ 2
...
```

right? And if instead of saying `do day1 day2 day3` which leads to exceptions for two and one day events we iterated over available days like


```
date:
  year: 2019
  month: May
  day
    - dom: 19
    main_event: {{ social }}
  day_ 2
    - dom: 20    
...
```

Either specifying a day of the week for each day, or specifying what day it starts on?

ericholscher:
I'm thinking something like:

```
days:
   - - dotw: Saturday
     - month: May
     - day: 18
     - summary: Foo summary
  -  - dotw: Sunday
      - month: May
```

So that we can iterate over it, and not have like "day_three" which breaks for AU, etc.

plaindocs:
Works for me.

plaindocs (approved):
Looks good.

plaindocs:
Shall we merge as is and tackle YAML stuff separately?

ericholscher:
Sure. 💯

PR #741, opened 2018-12-03
Hacky calendar with ICS export

plaindocs:
Proof of concept using @kelly-o 's numbers!

Could not be bothered making the day offsets relative in the end, would have had to invert the iteration and whatever else. :-p Happy to look at that if folks want, but...

- [x] add other info from the sheet (audience, links, desc)
- [x] ~configure notifications in advance (one week?)~ Just configure the default alarm of the cal we're adding it to :-D 
- [x] these are just generated dates, so we'd need to double check with the CFP page, I think I tweaked to Fridays or some such. 

image

[cal_and_dates.zip](https://github.com/writethedocs/www/files/2667286/cal_and_dates.zip)

--------

Update 18/01/19

- [ ] the sensible way to do this I think would be to only store the `yaml` cal data, and generate the ics with the build script 

Something like 

```
for each dir in confs/x
  make the ics
```

and tweaking `confcal ics` to take `source.yaml` and `output.ics` would be sufficient. In the Makefile even?


ericholscher:
Pretty neat. Agreed on the additional features before merging, but 💯

mxsasha:
Looks good! The Python developer in me does prefer not to use single-character variables as they are less obvious, and four-space indentation.

plaindocs:
Yeah, I can't really claim to be a python developer :-p

This got a little more involved when I started thinking about how to make the dates editable before writing the ical... Will hopefully push something up today.

plaindocs:
[_scripts.zip](https://github.com/writethedocs/www/files/2667264/_scripts.zip)

plaindocs:
And I think that is good to go.

Shall we we just check the yaml and ics into the repo? No big deal either way.

docs/_scripts/confcal.py
```diff
@@ -0,0 +1,70 @@
...
+    "Edit that file if you want to tweak the dates before creating the ICS"
+
+    d = Delorean()
+    d = parse(date, dayfirst=True)
+
+    test = dict()
+    for name, info in events.items():
+        date=(d-timedelta(days=info['notice'])).format_datetime('dd MMMM yyyy', locale='en_GB')
+        print(name, 'on', date )
+        test.update({name:{'date': date, 'audience': info['audience']}})
```
plaindocs, line 46: There is probably a better way to do this?

plaindocs:
And I guess we retire the spreadsheet, and add this to the org docs. 

And I could add the links to the event description or just make sure those are clear in the docs.

thatdocslady:
This is way too devvy for me to give proper feedback, so I trust y'all. The only thing I'm wondering (dreaming?) is possibly a way to automate Trello due dates around this, but it might be a distant dream...

ericholscher:
> Turn off CFP

Was this meant to go here? I guess we should ship it :)

plaindocs:
This is no longer relevant.

PR #753, opened 2018-12-17
ger-munich.yaml

kaiva:
I would like to add a new meetup group in Munich to the list of current meetups.


mjang (approved):
Works for me. Thanks so much for organizing a Meetup in Munich! Adding Sam Wright to the review for his info.

PR #754, opened 2018-12-18
Fix green header for Prague 2019.

mxsasha:
new screenshot:
screenshot 2018-12-18 at 16 51 28


plaindocs (approved):
Looks good.

We're gonna have some more of this type of thing here and for Lithuana, because some of Eric's fixes applied to Portland 2019 not the 2018 series I think. @ericholscher ?

ericholscher:
Yea. Most of the other changes I believe were in the templates, which should automatically be shared across confs. Hopefully we'll get more and more shared to reduce this issue :)

PR #756, opened 2018-12-19
Merging update to #job-posts-only purpose

MaggieFero:
New users have been confused by the description of #job-posts-only here, so I updated it to hopefully reduce that confusion.


ericholscher (approved):
Thanks!

PR #761, opened 2019-01-05
Portland 2019 CFP reminder and ticket sales blog post

thatdocslady:
Specific review points:

- [x] @ericholscher - I created the Tito event and updated ticket prices, but we need to enter GDPR information before we go live, and I don't know all of the fields (it's in Customize > Data Protection)

- [x] @plaindocs - I think we should probably extend the CFP by 1-2 weeks, and I noted this in the blog post (added 2 weeks but please tell me if that's too long, and I'll shorten it by however much you say)

- [x] I also don't know how to turn on ticket sale flags etc, if someone can push this to the PR it'll be great (and show me sometime how to do this?)

PR #762, opened 2019-01-05
Only show tickets on sale after the CFP ends

plaindocs:
Fix button display issue in #761 

An alternative would be to flip the logic, and not show the CFP there once tickets are on sale, as we've already got the *speak* button in the menu


thatdocslady:
I'm leaning more towards keeping the CFP button visible until it closes, and then switch to the tickets button. Seems to me like it's better to highlight the CFP while it's open, people can buy tickets later ;)

PR #765, opened 2019-01-11
Create an atom feed with absolute links for Mailchimp

ericholscher:
This will fix the issue where we have relative links in our emails.
Most feed readers do this automatically,
but not mailchimp.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -34,7 +34,7 @@ on_netlify = os.environ.get('BUILD_VIDEOS') == 'True'
 on_travis = os.environ.get('TRAVIS') == 'True'
 if not on_rtd and not on_netlify and not on_travis:
     exclude_patterns.append('videos')
-REWRITE_FEED = True
+REWRITE_FEED = False
 
 extensions = [
     'ablog',
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,8 +19,7 @@ from _ext.core import (
     set_html_context, unset_html_context
 )
 from _ext.meetups import MeetupListing
-#from _ext.videos import main
-
+from _ext.atom_absolute import rewrite_atom_feed
 
 exclude_patterns = [
     '_build',
@@ -35,6 +34,7 @@ on_netlify = os.environ.get('BUILD_VIDEOS') == 'True'
 on_travis = os.environ.get('TRAVIS') == 'True'
 if not on_rtd and not on_netlify and not on_travis:
     exclude_patterns.append('videos')
+REWRITE_FEED = True
 
 extensions = [
     'ablog',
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -26,12 +28,12 @@
 ]
 
 # Only build the videos on production, to speed up dev
-import os
 on_rtd = os.environ.get('READTHEDOCS') == 'True'
 on_netlify = os.environ.get('BUILD_VIDEOS') == 'True'
 on_travis = os.environ.get('TRAVIS') == 'True'
 if not on_rtd and not on_netlify and not on_travis:
     exclude_patterns.append('videos')
+REWRITE_FEED = False
```
plaindocs, line 37: Setting to `True` locally I get

```
File "/home/samuel/Nosync/Repositories/WTD/www/docs/_ext/atom_absolute.py", line 37, in rewrite_atom_feed
    namespace = root.nsmap.values()[0]
TypeError: 'dict_values' object does not support indexing
```

plaindocs (changes requested):
Same error on Netlify. 

I tried wrapping with `list()` but that doesn't fix it.

docs/conf.py
ericholscher: Hrm, probably a py3 issue, will poke at it.

plaindocs:
Don't think this works:

```
&lt;li&gt;250,000 sessions (up from 225,000)&lt;/li&gt;
&lt;li&gt;450,000 page views (same as last year)&lt;/li&gt;
&lt;li&gt;&lt;a class=&quot;reference internal&quot; href=&quot;../../guide/&quot;&gt;&lt;span class=&quot;doc&quot;&gt;guide&lt;/span&gt;&lt;/a&gt;: 225,000 views&lt;/li&gt;
&lt;li&gt;&lt;a class=&quot;reference internal&quot; href=&quot;../../conf/&quot;&gt;&lt;span class=&quot;doc&quot;&gt;conference sites&lt;/span&gt;&lt;/a&gt;: 120,000 views&lt;/li&gt;
&lt;/ul&gt;
```

Also, mother of GOD, that escaped atom feed is !!!

ericholscher:
> Also, mother of GOD, that escaped atom feed is !!!

Not sure I follow? The [feed](https://deploy-preview-765--dentist-canary-64055.netlify.com/blog/archive/tag/newsletter/atom_absolute.xml) I'm looking at looks great, with absolute links where previously they were relative?

plaindocs (approved):
Sorry, missed the "don't rewrite feeds locally", so was testing and getting relative feeds.

Also, probably we don't need that logic in? Could just rewrite to absolutely everywhere.

ericholscher:
> Also, probably we don't need that logic in? Could just rewrite to absolutely everywhere.

This is mainly because of the lxml dependency. It can be a PITA to install, but if we don't care about that, I can remove the logic.

plaindocs:
I don't feel very strongly about it, but also wasn't aware of the dependency issue.

ericholscher:
Will ship this for now, and we can futz with it in the future if needed. 👍

PR #769, opened 2019-01-15
Upgrade our sphinx version & deps

ericholscher:


requirements.txt
```diff
@@ -1,6 +1,6 @@
-Sphinx==1.6.7
-recommonmark==0.4.0
-ablog==0.9.2
+Sphinx<2
+recommonmark<1
```
mxsasha, line 2: Why don't we pin the versions down? I'm not too familiar with this package, but it seems the upgrade from what we had, within the same major version, required code changes? So not something we would want to have randomly happen.

mxsasha:
Per my inline comment, how about we pin specific versions and use something like pyup.io ?
It creates PRs for version updates, often with changelogs right in the PR, causing CI to be run with the new dependencies. So it means we can stay up to date, but also not have surprise version upgrades during a deploy. Free for open source.

https://github.com/irrdnet/irrd4/pull/159 is an example of a PR pyup.io generated.

ericholscher:
That makes a lot of sense. Another option is this tool we use on RTD that does about the same thing, but manually: https://github.com/rtfd/common/blob/master/tasks.py#L301 -- if we want to setup the service tho I'm all for it.

mxsasha:
I'm going to try to setup pyup.io. I really like the single-click upgrades.

mxsasha:
See #771 and #772.

ericholscher:
Closing in favor of fancy new stuff 👍

PR #770, opened 2019-01-15
Add job board to the site sidebar

ericholscher:
This also includes our upgraded deps PR


mxsasha:
Why is #769 part of this PR again too? Because #772 fixed that already, so now it conflicts?

ericholscher:
@mxsasha was just fixing it :)

plaindocs (commented):
A++ great stuff.

BUT OMG that hanging button is too ugly to publish (am I channeling Mikey here?).

I looked at 

```
+                   .attr('class','toctree-l1')
```

but the first level li in the navbar doesn't have bullets.

We could move the line about the board to `l1` and make all of the jobs`l2` to mirror the nav display?

ericholscher:
I futzed with the styles, we can remove the list style, but then it's hard to tell the listings apart.

ericholscher:
@writethedocs/core Review please? Would like to get this shipped 👍

PR #778, opened 2019-01-21
Update testing for vale

mjang:
As Vale includes write-good, I think we can delete the section for that.

I could go into more detail w/r/t Vale installation, but that would seem too far beyond the other entries on this page.

I've also removed the author/date entries to encourage input from others.


ericholscher (approved):
Looks good. 👍

PR #793, opened 2019-02-24
vilnius update

aidiss:
Updated some information like conference venue, dates, duration, organizers and etc.
Can't guarantee that everything is in place as there is massive amount of files.

Let's start minimum required info please.

### DIFF
--- a/docs/conf/vilnius/2019/about.rst
+++ b/docs/conf/vilnius/2019/about.rst
@@ -4,7 +4,7 @@
 About the Conference
 ====================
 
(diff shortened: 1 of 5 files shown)


plaindocs:
This is a good start! 

The main thing you need to look at is `docs/_data/config-vilnius-2019.yml` and add ticket prices.

@thatdocslady are you firing up a tito setup?

thatdocslady:
We're only publishing the landing page this week, so no need for Tito site yet. That will be a part of the CFP/ticket/full website push in 2 weeks.

PR #794, opened 2019-02-25
Update ticket prices

aidiss:


docs/_data/config-vilnius-2019.yaml
```diff
@@ -16,13 +16,13 @@ photos:
 
 tickets:
   community:
-    price: $750 # TODO
+    price: $150 # TODO
```
plaindocs, line 19: € right? 

And can you delete the #TODOs as well?

plaindocs (commented):
Otherwise looks good

PR #807, opened 2019-03-01
Switch to using explicit lists for callouts instead of crazy conditionals

ericholscher:


plaindocs:
Nice. Thing is we then have to remember to update the CTAs as well as the Flags, which was the attraction of the CCs.

Any thoughts on how we can have the best of both worlds?

plaindocs:
And we should include Vilnius if we're updating 2019 anything so we're on the same page.

ericholscher:
> Any thoughts on how we can have the best of both worlds?

It just seems like there isn't actual state that we can capture here in a meaningful way. There's just too many times we want different things to be highlighted, depending on eg. sponsorship sales, ticket sales, etc. Seems like explicit is best here, so we can just make it say what we want when we want.

ericholscher:
@plaindocs Updated Vilnius -- shall we merge and see how this goes, and we can move back towards conditionals if we don't like it?

plaindocs:
Sure lets do it. 

Merging `master` will fix the build. 

We also have a PR open for Australia,  dunno if you wanna bother applying there also.

ericholscher:
Yea, I want to do something similar to this for dates instead of `day_1`, because I think it will work much better in all cases. But baby steps :)

plaindocs:
That's odd.

plaindocs:
Why didn't the YAML linter pick up on that?

plaindocs:
lol it did https://travis-ci.org/writethedocs/www/jobs/509400493

ericholscher:
:+1:?

plaindocs:
:ship::it:

plaindocs (approved):

PR #824, opened 2019-03-14
Fix duplicate talk abstract

bethaitman:
Was going through the old talks to put together an index and I noticed that this abstract was duplicated, so replaced it with the correct one.

I had a stab at formatting the data in the right way - hopefully it's worked!


bethaitman:
Ok, looks like I screwed up the yaml. Will come back and fix it when I get the time.

plaindocs:
@bethaitman looks great to me! Thanks!

plaindocs (approved):

PR #841, opened 2019-03-21
Newsletter link checking

plaindocs:


ericholscher:
Looks like this is already working since we're excluding old job postings already? 🎉 /cc @bethaitman

plaindocs:
I KNOW RIGHT? :sun_with_face:

ericholscher (approved):

ericholscher:
I imagine this will start randomly failing over  time as URL's we link to disappear, but that's good to know at least -- perhaps we can just point to archive.org copies when that happens.

ericholscher:
🚢

PR #849, opened 2019-03-25
Better link to google maps for MeMenamins

timothymcmackin:
The old link takes me to a disambiguation-type page with a few options for three different "Crystal Ballrooms".


plaindocs (approved):
Awesome thanks!

PR #851, opened 2019-03-27
Update Netlify logo

verythorough:
The Netlify logo has changed a little bit, so this updates it. I re-sized to match the one that was already in here.


ericholscher (approved):
Thanks! 👍

PR #861, opened 2019-04-03
and I found a last minute-typo

bethaitman:


bethaitman:
one character change, thought I could probably just merge

PR #867, opened 2019-04-05
Added TC Camp (4/26-27) to Community events list

clairealyssa:
Completely forgot about this pre-publish. Could we get this added in @bethaitman ?


plaindocs:
So the list isn't really community events, but WTD meetups. Do we want another section for external events? Which would be another maintenance issue.

bethaitman (approved):
Sorry, I didn't see this over the weekend! @plaindocs, I think we talked about this a month ago in the newsletter channel and figured it was close enough to the community that it was okay.

clairealyssa:
I'll keep that in mind in the future.

PR #873, opened 2019-04-14
writing/mindshare: Reword a sentence

cfra:
Reading through the guide I had to read this sentence multiple times
to parse it correctly.

In my humble opinion, it is better to avoid splitting "should be as
easy and straight-forward as possible" by placing the subject of
the sentance in its middle.


plaindocs (approved):
Nice!

PR #878, opened 2019-04-27
Update speakers

plaindocs:
- [x] Figure out what to do with the schedule before merging


ericholscher (approved):

docs/_data/2019.portland.speakers.yaml
```diff
@@ -302,29 +302,6 @@
...
-  speakers:
-  - name: 'Syne Mitchell'
-    slug: syne-mitchell
-    twitter: synemitchell
-    website: https://www.synemitchell.com/   http://www.linkedin.com/in/allen-tsai-ux-lead
-  - name: ' Allen Tsai'
-    slug: allen-tsai
-    twitter: synemitchell
-    website: https://www.synemitchell.com/   http://www.linkedin.com/in/allen-tsai-ux-lead
-  title: Bam! The Power of Comics to Explain Technology
```
ericholscher, line 327: Should we keep this on the listing, and put a `CANCELLED` in front of it? I'm torn as to what is best.

ericholscher (commented):
Need to fix the old `News` links, which makes me think maybe we should keep the content around and just update that it was cancelled.

docs/_data/2019.portland.speakers.yaml
plaindocs: That would be the advantage up generating the link in old news with templates. Unless you *really* value the archive factor of old content.

plaindocs: The only breaking broken link (:-p) is the schedule, because I've not updated that.

ericholscher: We already sent out an email with the old content that we can't change.

plaindocs: Yeah, but I don't think that is a huge concern. Let's discuss on slack?

ericholscher:
I think the best option for the schedule is to just move the last speaker Monday into that slot.

ericholscher:
I think the best option for the schedule is to just move the last speaker Monday into that slot.

docs/_data/2019.portland.speakers.yaml
ericholscher: Yea, let's just go ahead and remove it then 👍

docs/_data/portland-2019-day-1.yaml
```diff
@@ -68,16 +68,9 @@
...
-  Slug: mark-f-iverson
-  Time: '4:50'
 - Duration: '10'
   Session: Group Photo
-  Time: '5:20'
+  Time: '4:40'
 - Duration: '0'
   Session: Day 1 wraps up
-  Time: '5:30'
+  Time: '4:50'
```
ericholscher:
```suggestion
  Time: '5:00'
```

Just to make it round out nicely.

ericholscher (approved):
Looks good. We should ping the speaker we moved to just confirm the change.

PR #896, opened 2019-05-12
Add pic of Leona to WW page for 2019 conference

canncrochet:
Leona is stepping in for Tana for the WW. Tana has agreed to keep her pic on the WW page, in the event that someone has a question for WTD crew, and they recognize her from the page.


ericholscher (approved):

ericholscher:
Thanks for the update!

PR #905, opened 2019-05-16
Add livestream to nav and index for Portland 2019

mxsasha:


ericholscher (approved):
Looks good, thanks! We should also add it to the nav bar in another PR, so we have links to it for users in the future.

mxsasha:
I think it's already in the nav bar? First link under program, triggered by setting livestream to True.

ericholscher:
Ah, neat! We should probably include it in the nav unconditionally, so people know it's gonna happen.

PR #907, opened 2019-05-16
Fix link syntax

verythorough:
Forgot the `_` at the end of one link.


verythorough:
Oops, still broken. Fixing.

verythorough:
Works now. :)

PR #908, opened 2019-05-19
Add GitLab to the Portland 2019 Writing Day page

axilleas:
:wave: Here's a bit of info about GitLab and what people should expect at Writing Day.


mjang (approved):
with recent typo correction

PR #917, opened 2019-05-22
[DRAFT] Add Vilnius 2019 schedule

mxsasha:
The day does end quite early (schedule page still says 18:00, we need to update this if this is the final schedule).
The original planned time was from 10-18, but we have 3 hours of talks, 40 minutes of lightning talks, an hour of lunch, an hour of snack breaks, ~30 minutes of speaker switching. So that's just over 6 hours - we just can't really fill 8 hours, hence the early ending.

Also, please double check that I didn't make any timing calculation errors.
(snack breaks are 30 minutes on purpose for a bit more padding)

### DIFF
--- a/docs/_data/config-vilnius-2019.yaml
+++ b/docs/_data/config-vilnius-2019.yaml
@@ -22,8 +22,8 @@ buttons:
     #link: /cfp
     - text: Buy a Ticket!
       link: /tickets
-    - text: See the talks!
-      link: /speakers
+    - text: See the Schedule!
+      link: /schedule
     # - text: Prepare with the Welcome Wagon
     #  link: /welcome-wagon
     # - text: Watch the Live Stream
--- a/docs/_data/config-vilnius-2019.yaml
+++ b/docs/_data/config-vilnius-2019.yaml
@@ -84,7 +84,7 @@ cfp:
 
 flagspeakersannounced: True
 flagcfp: False
-flaghasschedule: False
+flaghasschedule: True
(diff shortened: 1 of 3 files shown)


thatdocslady:
Thanks @mxsasha looks good!

I'm wondering if extending the snack breaks might not be too much for an already-stretched schedule. Maybe instead, we can shrink them back and add 1-2 hours of post-conference networking time (let's say 16:00-18:00) in lieu of the welcome reception that we're not formally hosting. 

Thoughts?

mxsasha:
Yes, that sounds good. We can also cut lightning talks to 30 minutes (now 40). That saves us a total of half an hour, ending at 16:10, and then networking time until 18:00? Or 17:30.
(needs a better name for the schedule probably)

thatdocslady:
@mxsasha great! Let's do that :) I say ending at 17:30 is fine, but maybe 18:00 looks better on the schedule? As for the name, how about something like "Post-conference social time" or "Informal networking"... I'll think about it some more :)

mxsasha:
I added it for now as informal networking until 18:00. Someone should double-check I updated all the start times correctly, snacks are 20 minutes now, lightning talks 30.

aidiss:
I agree regarding post-conference social time.
Do you need anything from me?

aidiss (approved):
Looks good

thatdocslady:
Thanks @aidiss, I'll merge this PR now and start working on the blog post.

PR #933, opened 2019-05-30
Add link to Mike Jang's slides

mjang:
Not sure how to set up a clickable link in yaml, but maybe it's not important


plaindocs:
The better way to do this would be to add a `slides:` key like in 2017. 

But that probably requires a modicum of work to display. I'll try to get to that on Sunday.

mjang:
Updated with `slides` slug. I'm not sure whether order matters.

PR #960, opened 2019-06-02
WIP: integrated sphinxcontrib-spelling

sergeykolosov:
Things to consider:
* spelling word-list is pretty much empty yet;
* it may be worth using a separate tox environment for spelling (there are both pros and cons).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ install:
  - eval "$(gimme 1.8)"
  - go get github.com/errata-ai/vale
  - sudo apt-get install libcurl3-dev
+ - sudo apt-get install enchant
 # Safelist all Greenkeeper.io branches https://greenkeeper.io/faq.html
 branches:
   only:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -41,7 +41,9 @@ REWRITE_FEED = False
 
 extensions = [
     'ablog',
+    'sphinx.ext.intersphinx',
     'sphinxcontrib.datatemplates',
+    'sphinxcontrib.spelling',
     'notfound.extension',
     'recommonmark',
 ]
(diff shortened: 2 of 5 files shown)


plaindocs:
Cool stuff! Got a link to pros and cons of the separate environment? ;-)

plaindocs:
Hey @mxsasha @ericholscher did you see this? Any thoughts on whether it is worth doing the initial work of populating the "correct words list"?

plaindocs:
I'll find out what amount of words we're talking about here.

plaindocs:
After kicking some deps I found

WARNING: Found 2864 misspelled words
build succeeded, 6 warnings.

Which then boils down to 660 words, mostly names and brands. 

[spelling-unique.txt](https://github.com/writethedocs/www/files/3376328/spelling-unique.txt)

Would be pretty straightforward  zoom through that list, and add it directly to the spelling file.

plaindocs:
One could even easily script adding all speaker names to the list when adding a speaker YAML.

Plus some decent config option https://github.com/sphinx-contrib/spelling/blob/master/docs/source/customize.rst#word-filters

Overall I'm inclined to add this in. @mxsasha @sergeykolosov any more thoughts?

mxsasha:
I think this is good! Maybe we shouldn't add all the misspelled current words though? I see a bunch of British spelling in there (this may have been me 😬).

plaindocs:
I'll pick this up, add some  words (trying to avoid British spellings)

plaindocs:
Replaced by #1554 

I ran into issues with `ablog`, and we're already using `vale` elsewhere. If we ever do switch back to this the word list is reusable anyway, and that is where most of the work is.

PR #963, opened 2019-06-02
Fixed missing image src attributes

bandantonio:


osg (commented):
Where / How can one QA this? (For example, a computer will not know if file Chris was accidentally named Beth and vice versa.

plaindocs (approved):

plaindocs:
Thanks @bandantonio 

@osg all it is doing it saying that for name `X` try using image `x.jpg`. Yeah, the website code can be clunky in places.

PR #973, opened 2019-06-07
Decouple meetups and videos

plaindocs:
So RTD actually does generate the meetup events.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -134,9 +134,10 @@ html_context = {
 
 if build_videos:
     from _ext.videos import main
-    from _ext.meetup_events import main as meetup_main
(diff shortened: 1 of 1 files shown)


ericholscher:
I don't understand, can't we just set the variables on RTD?

ericholscher:
We don't want all users to have to build the meetup events whenever they are building the docs locally or contributing. We'll need to make the code a bit more defensive (only running when the `API_KEY` env var is there) to make this run by default I think.

plaindocs:
Right, that makes sense.

I was confused by the intersection of on_rtd AND build_video above but only the build_video below. 

Adding the video var to RTD makes that work.

PR #463, opened 2018-03-21
Generate upcoming meetup events

plaindocs:
This is a less awful replacement of #254 that uses the API.

Caveats:

- [x] you need an API key (I'll add it to Netlfy/Whatever as a secure variable at some point) export in bash or so like `export MEETUP_API_KEY=`
- [x] Ugly Meetup IDs -> should probably keep the ID in the YAML instead of the URl if we're using it like this
- [x] We should be able to run it on cron or something eventually
- [x] Need to actually write the RST / YAML / whatever

Need to clean up the output, handle the dates etc. 

POC

```
'##Write the Docs - San Francisco Bay Area'
'(SF) Hack-A-Thon! Swimming in the deep water: a lone writer’s survival guide'
'https://www.meetup.com/Write-the-Docs-SF/events/248343809/'
'##Write the Docs SLC'
"April Meeting - Creating customer support content when you don't have support"
'https://www.meetup.com/Write-the-Docs-SLC/events/248894227/'
```

And the dataformat it returns is something along the lines of:

```
'created': 1519883561000,
  'description': "<p>• What we'll do<br/>If you need help setting up a writing "
                 'portfolio, this Meetup is for you!</p> <p>Erin Grace has '
                 'agreed to preview her talk for the Write the Docs Conference '
                 "in May. Here's her abstract:</p> "
                 '<p>***********************<br/>In a tech writer’s career, '
                 'creating and maintaining a great portfolio is as important '
                 'as it is stressful. What belongs in there? What constitutes '
                 'your "best work"? What if all your work is proprietary? Do '
                 'you really need a portfolio if you have a great job?</p> '
                 '<p>Building a portfolio is an exercise in documenting '
                 'yourself: your writing, your skill, your journey as a '
                 "professional. Like any other kind of documentation, it isn't "
                 'easy or intuitive to create something great. And like any '
                 "other kind of documentation, it's worth doing the hard work "
                 'to get it right.</p> <p>In this talk, I cover:</p> <p>* Why '
                 'you need a portfolio even if you have a stable job<br/>* How '
                 "to create a well-rounded portfolio and decide what's your "
                 'best work<br/>* Strategies for adding proprietary work '
                 "(legally!)<br/>* Options to present it that aren't just a "
                 'series of email attachments</p> '
                 '<p>***********************</p> <p>• Important to '
                 "know:<br/>Please familiarize yourself with Write The Docs' "
                 'Code of Conduct, as it applies to this Write The Docs '
                 'PDX.</p> <p><a '
[...]

PR #976, opened 2019-06-09
Fix meetup build error

plaindocs:
I'm not really sure why the build started failing soon after I merged #463. Investigating. 

https://travis-ci.org/writethedocs/www/builds/543128866 

@mxsasha any idea?

I restarted the last successful build https://travis-ci.org/writethedocs/www/builds/542638416 and it now errors, so might be a dependency somewhere?


mxsasha:
The traceback is a bit confusing, because the issue does not actually seem to be with the `docs/conf.py` file that `eval_config_file` is attempting to read. The code seems to be trying to decode an empty JSON response, and therefore my best guess is a flaky meetup API. This is consistent with the fact that it almost always fails locally for me, at a different point when getting meetup info. When I dump the response content by hacking `meetup/api.py`, I do see server response headers, with `Content-Length: 0` and `Content-Type: application/json;charset=utf-8`, which matches an entirely empty JSON response, which I think is indeed not valid.

In other words, best guess is an extremely flaky meetup API, with a possible hacky solution being to have `meetup_events.py:49` simply retry if a `JSONDecodeError` is raised. I can't assess immediately whether a single call is made per meetup or one for all meetups, because in the latter case it'll help poorly. Perhaps it's easy to call the API once per meetup instead, so that we can implement the retry in our own code. Or, there may be an entirely different solution.

plaindocs:
I reckon you're right. Seems intermittent. 
I'll build some error protection in so failed meetups doesn't break the rest of the build.

plaindocs:
Urgh https://travis-ci.org/writethedocs/www/jobs/544273361

mxsasha:
> Urgh https://travis-ci.org/writethedocs/www/jobs/544273361

Wait, that's an entirely different issue, right?

plaindocs:
I think we can just close this.

PR #978, opened 2019-06-11
Be a little more careful with Meetups code

plaindocs:


mxsasha (approved):
This looks sensible in general, though is it also meant to fix #976?

plaindocs:
Well, I realized that builds by people who aren't members of the the wtd org get failing builds because they don't have access to the API KEY.

And it 'kind of' fixes intermittent API issues by not failing if it can't. Really I should tweak it to at least write a file that says no meetups due to connectivity issues. :-p

But feel free to improve or suggest improvements. Python isn't my strong point.

mxsasha:
Well, let's go ahead with this, because it should be able to run without the API key set for local work anyways. I will try to have a look at #976.

mxsasha:
It seems to consistently work for me nowadays, so perhaps this was just a temporary fluke in the meetup API? Let's not work more on #976 unless it happens more often.

PR #987, opened 2019-06-19
Update style guide

ericholscher:
Created a new style guide that covers some more areas


docs/style-guide.rst
```diff
@@ -1,10 +1,45 @@
+-----------
 Style Guide
-===========
+-----------
+
+This document specifies how we write on our websites.
+It's just a start,
+but everything has to start somewhere :)
+
+Part 1: Technical Formatting
```
plaindocs: There is no reason for these to be manually numbered.

plaindocs: Unless I'm missing something?

ericholscher: I think mostly to be able to reference them, but probably not.

Geodun: Yeah, numbering was for easy referencing. The Chicago Manual of Style uses it, and it's one of features I like about it. But it might not be as useful for style guide that's this small.

ericholscher:
Merged, this is way better than what we had :+1:

PR #991, opened 2019-06-23
Added a page introducing the XML markup language

ghost:
I wrote a page introducing the XML markup language. Similarly to the other markup pages, I included a section on what it is, why use it, how to use it, and a couple resources on XML. This page is written in reST.


plaindocs:
Thanks for this! I'll check and merge, but for now you'll want to add the filename (without extension) to the `toctree` around line 54 of `guide/writing/index.rst` and delete the related place holder below it.

docs/guide/writing/xml.rst
```diff
@@ -0,0 +1,63 @@
+===================
+Introduction to XML
+===================
+
+What is XML?
+============
+
+Similar to reStructuredText, Markdown, and HTML, XML is a markup language. Unlike other markup languages, it is not predefined and you must define your own tags and attributes. It's different from other markup languages in that instead of being used to present content on a page, it distributes that page's data over the Internet.
```
plaindocs: Not 100% sure I personally agree with some of these definitions, but I think we're ok to merge and iterate on this.

@writethedocs/core any objections?

plaindocs (approved):

docs/guide/writing/xml.rst
bandantonio: I would rewrite this section completely. First of all, it is not clear what XML actually is and why should I use it. Second, the section is too wordy (markup language everywhere) and there are styling inconsistencies (e.g. it is and it's).

ghost: > 
> 
> I would rewrite this section completely. First of all, it is not clear what XML actually is and why should I use it. Second, the section is too wordy (markup language everywhere) and there are styling inconsistencies (e.g. it is and it's).

I can make a quick edit on this section :)

bandantonio: > I can make a quick edit on this section :)

That would be great! I will post my suggestions shortly

bandantonio: I would rewrite this section as follows:

XML is a meta markup language developed for the Internet. It is derived from SGML (Standard Generalized Markup Language), the mother of all markup languages. XML was designed to store and transport data. Unlike HTML that defines how data looks, XML defines what data is. Users can use XML "as is" or adapt it to their needs by defining new tags or building blocks.

docs/guide/writing/xml.rst
```diff
@@ -0,0 +1,63 @@
...
+
+XML defines a set of rules to encode data in a format that is both human-readable and understandable to computers.
```
bandantonio: XML defines a set of rules to encode data in a format that is both human and machine readable. That is why XML syntax is used in Microsoft Office, Apple iWork, LibreOffice, RSS, SVG, and many other tools.

docs/guide/writing/xml.rst
```diff
@@ -0,0 +1,63 @@
...
+============
+
+Because you define your own tags and attributes, you can use XML to create essentially your own markup language. Defining these unique tags and attributes also gives structure to your document.
+
+Additionally, XML stores data in plain text. This makes it independent on the software and hardware to which the data wrapped in XML is being sent, received, and stored. This design also makes it easier to update to new browsers, new operating systems, and new software without losing data. 
+
+How To Use XML
+==============
+
+Let's say we wanted to create a table laying out the trees in a certain region. We could define a <tree> tag that would have the names of the trees, a <height> tag, a <type> tag, and more. These tags give structure to our document and it's fairly easily readable to humans.::
```
bandantonio: 1. we wanted -> we want
2. remove noise: fairly easily. Just write: and it's still human readable

docs/guide/writing/xml.rst
```diff
@@ -0,0 +1,63 @@
...
+
+    <tree>
+        <name> Black Walnut </name>
+        <height units="m"> 16 </height>
+        <type> Deciduous </type>
+    </tree>
+    
+XML is often used in concert with HTML, CSS, and JavaScript.
+
+Displaying XML
```
bandantonio: Inconsistent heading. Please change to `How to display XML`

bandantonio (changes requested):
I can make the requested changes on my own if you want to

docs/guide/writing/xml.rst
ghost: Thanks for the suggestions! I added them to the xml file. Hopefully I'm using GitHub right. Feel free to make any other changes and additions too.

bandantonio: Seems to be ok for me now. Thanks for quick changes!

bandantonio (approved):

PR #993, opened 2019-06-24
Updates to Documentation Guide

### LINKED ISSUE #990: Change the title of the Mardown page to "Introduction to Markdown"
For consistency, I suggest changing the title of the [Markdown](https://www.writethedocs.org/guide/writing/Markdown_basics/) page to "Introduction to Markdown" to match the other markup language page titles. 

Alternatively, change the page titles to "[Markup language] Basics."

I hope this isn't too nit-picky :)

bandantonio:
* Made markdown guide page title consistent with other pages.
* Removed incorrectly rendered headers from TOC

Before | After
-------|------
before | after

Fixes #990


plaindocs (approved):
Thanks @bandantonio ! :+1:

plaindocs:
Although now I come to think of it, something like 

```
.. raw:: html

   <h3>This is a third-tier header</h3>
```

might be a better approach? What do you think?

bandantonio:
> might be a better approach? What do you think?

Yes, it might be. Initially I thought about it but my approach was way too complex. Let me check your suggestion.

P.S: I was managed to run docs locally. I was missing this so much in Vilnius during the Writing Day :)

plaindocs:
Yeah, it makes stuff *much* faster! But takes quite some bandwidth and fiddling to get it to work. Not so easy when there are 10 people around a table doing different things.

bandantonio:
Just spotted that this is the only page in guides that has `.md` extension :grin: I think we should merge this PR 'as is' and then I will create another issue to rewrite markdown guide using reST

plaindocs:
There is no real need to rewrite that page in RST, even if it is the only one.

PR #1003, opened 2019-07-02
[WIP] Clean up Python code, add some documentation.

mxsasha:
Notable changes:
- Removed a bunch of code that has not been used for a long time, including unused jinja filters (perhaps most significantly: removed `conferences` from the default template context, which I'm fairly sure is unused)
- Improved detection and error reporting for rendering conference pages. This stops the "Unable to process conference YAML data: conf/index" warnings and provides more useful messages in cases of actual errors.
- Split and moved some things into different files to simplify core.py
- Moved all scripts (as in, manually run things) to _scripts.
- Removed a number of obsolete files
- Added docstrings to virtually all Python code
- Renamed quite a few things to be more clear on their purpose/scope
- Extended documentation somewhat to point more in the right direction
- Cleaned up for pep8

I've left some of the video stuff a bit alone, as it warrants a more thorough look specifically at video.

I've verified that `_build/html` produces identical output locally (with building videos enabled), so it looks like nothing broke.


mxsasha:
(Netlify failure may be due to cloudflare being broken right now)

plaindocs:
This looks awesome. I've skimmed the changes rather than dig into them though.

ericholscher (approved):
Lots of good stuff here! <3

PR #1022, opened 2019-07-19
Interview rachael stavchansky

Geodun:
adding interview with Rachael Stavchansky to hiring guide


Geodun:
I've been using  this syntax for external links: [ .. _Python: http://www.python.org/ ]. But it doesn't work well with multiple words. Is there a better way to attach an external link to multiple words?

Geodun:
Also, are content tags delimited by commas?

ericholscher:
@Geodun https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#hyperlinks

ericholscher:
We should keep tags as URL slugs, so lowercase and hypens in between words.

PR #1023, opened 2019-07-19
Interview alex ball

Geodun:
adding interview with Alex Ball to hiring guide


Geodun:
These interviews are going to be coming in fairly frequently over the next couple weeks, and that will cause some conflicts to the index. Should I hold off on updating the index until the rate of additions cools off a bit?

ericholscher:
@Geodun we could use the `:glob:` setting on the toctree and just do a `*` which will automatically include all of them. Especially if we don't care about ordering.

PR #1058, opened 2019-08-21
Add hoverxref extension for testing

ericholscher:


humitos:
@ericholscher can you please re-trigger a build for this PR? The tooltip should display properly now. Let me know.

humitos:
Example building locally:

Screenshot_2019-08-22_15-04-19

humitos:
Just to note that there is nothing to change in this PR, new `master` branch in the extension has a small CSS fix for this.

humitos:
Looks awesome!

ericholscher:
Yea, I think we need to do some more additions around `:ref:`'s to make this really useful, we don't use them enough. It's an obvious win for using Sphinx directives over basic RST links tho 👍

plaindocs:
If we hand picked a list of 20 pages or so that we think are "important" we could get a writing day task to convert them to directives.

PR #1056, opened 2019-08-20
Update gulpfile.js to Gulp 4

plaindocs:
According to https://www.liquidlight.co.uk/blog/how-do-i-update-to-gulp-4/

Fixes #1042

PR #1066, opened 2019-08-30
Update gulpfile to gulp 4

plaindocs:
This looks like it might work.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -9,9 +9,9 @@ var gulp = require('gulp'),
     del = require('del'),
     plumber = require('gulp-plumber'),
     browserSync = require('browser-sync'),
-    changed = require('gulp-changed'),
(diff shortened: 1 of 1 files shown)


plaindocs:
Replaces #1056 and takes #1010 into account.

plaindocs:
Browsersync isn't working, but styles is.

mxsasha (approved):
works for me now

plaindocs:
https://github.com/gulpjs/gulp/blob/master/docs/recipes/minimal-browsersync-setup-with-gulp4.md

plaindocs:
I do wanna highlight that Browsersync still doesn't work here, so if we merge this we should open an issue to fix it. Pretty low priority though.

ericholscher:
This worked for me as well :+1:

PR #1077, opened 2019-09-13
Switch due to cancellation

plaindocs:
- [x] Confirmation pending


docs/_data/prague-2019-day-1.yaml
```diff
@@ -63,12 +63,9 @@
 - Duration: '10'
   Session: Switch Speakers
```
mxsasha: We don’t need the switch speakers there right?

ericholscher (approved):
I removed the last switch speakers, since we generally don't leave time for it and don't want to confuse the unconf folks.

PR #1087, opened 2019-09-15
Fix corilla link

hhoibraten:


plaindocs:
Thanks! We're trying to keep all links https if possible. I know Corilla doesn't have one. I''' figure this one out.

PR #1091, opened 2019-09-15
updated the cfp links

jmessias3:
I have deleted the previous years for cfp writethedocs/www/.tox/all_links/tmp/html/organizer-guide/confs/cfp/index.html and added the links for 2018 and 2019


plaindocs:
Awesome! I feel like we should be able to say year-1 and year -2 here, and then forget about this :-D

PR #1094, opened 2019-09-22
Update the link used to join slack

ericholscher:
This uses the new Slack invite link syntax,
instead of our old-school invite API.
They said the old API is deprecated so we need to move to this method.


plaindocs (approved):
:+1:

PR #1100, opened 2019-09-23
Limit speakers to twice in four years

plaindocs:
Any thoughts or opinions?

### DIFF
--- a/docs/conf/portland/2020/cfp.rst
+++ b/docs/conf/portland/2020/cfp.rst
@@ -43,6 +43,7 @@ We welcome talks from first-time speakers, from industry experts, and from every
 Whatever your background and experience, we prefer hearing about new approaches than about tried-and-tested technology.
 We especially welcome talks from underrepresented groups within the tech community.
 We want to hear a variety of viewpoints, so we limit speakers to two talks in any four year period at each location.
+If you're a WTD organizer, please only submit talks to conferences you're not actively organizing.
 
 **Mix of roles and perspectives**
 
--- a/docs/conf/portland/2020/cfp.rst
+++ b/docs/conf/portland/2020/cfp.rst
@@ -43,7 +43,7 @@ We welcome talks from first-time speakers, from industry experts, and from every
 Whatever your background and experience, we prefer hearing about new approaches than about tried-and-tested technology.
 We especially welcome talks from underrepresented groups within the tech community.
 We want to hear a variety of viewpoints, so we limit speakers to two talks in any four year period at each location.
-If you're a WTD organizer, please only submit talks to conferences you're not actively organizing.
+If you're a WTD conference organizer, please only submit talks to conferences you're not actively organizing.
 
 **Mix of roles and perspectives**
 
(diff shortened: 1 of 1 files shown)


mxsasha (approved):

Bradamant3:
Do we want to say anything, here or elsewhere, about limiting talks by organizers? I don't object to the talks we have had -- Mike's especially come to mind as appropriate and a good example of why we should (sometimes) include them. Ditto Neal's, although he's come to community helping mostly after the couple of brilliant talks he's given.

I've never submitted a talk to Portland bc I think it's probably not appropriate, and the longer I help with community work and running any conference, the more I think it's not a good idea for me to speak. But I'm not sure how other organizers might feel on the issue.

If we think such a limitation is a good idea, I recommend publicly documenting it to let the world know that we're working to keep every sort of opportunity as open as possible.

Bradamant3:
Whoops, didn't mean to only take the conversation in a different direction.

Explicit +1 to this PR.

bethaitman (approved):

docs/conf/portland/2020/cfp.rst
```diff
@@ -42,6 +42,8 @@ That said, there are some more helpful and concrete pointers below.
 We welcome talks from first-time speakers, from industry experts, and from everyone in between.
 Whatever your background and experience, we prefer hearing about new approaches than about tried-and-tested technology.
 We especially welcome talks from underrepresented groups within the tech community.
+We want to hear a variety of viewpoints, so we limit speakers to two talks in any four year period at each location.
+If you're a WTD organizer, please only submit talks to conferences you're not actively organizing.
```
mjang: I like the PR in general. I don't know if what I do qualifies me as an "organizer" -- or if it's based on those listed in links like http://www.writethedocs.org/conf/portland/2019/team/ . 

If it is, we might also want input from @canncrochet and @mopdx, as I'm guessing they might want to speak again in the future.

In addition, I might say "WTD conference organizer", to avoid confusion with WTD Meetup organizers.

mopdx: I like this idea. Not submitting talks to the conference which one is actively organizing helps with the optics, in addition to opening up slots for other participants. I think several things have changed since Write the Docs began its journey in 2013:

1. We had 1-2 yearly conferences (now there are at least 3-4 per year). So finding a WTD conference to submit a talk to is easier.
2. Most WTD conferences have unconference tracks, so if there is a burning desire to present, the unconfy might be a better venue.

My two cents' worth.

canncrochet: I think I'm late to the party, I apologize. This seems fair. It helps to protect conference organizers from over extending and encourages diversity in speakers, including first-time speakers. Looks good!

plaindocs: Thanks folks! Excellent feedback all. And certainly no apology needed @canncrochet. 

I think we have enough consensus and :+1: on this to merge it in. As always, edits and improvements welcome after merging as well as before!

PR #1104, opened 2019-09-28
Fix formatting for ordered lists on the "Introduction to Markdown" page

joncros:
(Changed lines 62-79 of markdown.md.) The example was rendering as items 1-5 without any indentation.
  1. Just type...
  2. If you want...
  3. If you make...
  4. If you press...
  5. If you want...
  - If need be,...

Fix example so that indents render and the numbering restarts on the fourth and fifth list items, as intended.

(Note that the github file viewer substitutes 'i' for '1' on a list item indented once, and it substitues 'a' for '1' on a list item indented twice. But the html output by sphinx restarts the numbering at '1' as intended.)


plaindocs (approved):
Well spotted! Thanks!

PR #1128, opened 2019-11-01
vale changes install again

plaindocs:
Should fix the latest broken builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ before_install:
 install:
  - pip install tox-travis
  - eval "$(gimme 1.9)"
- - go get github.com/errata-ai/vale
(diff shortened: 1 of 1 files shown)


ericholscher (approved):
Annoying :/

jdkato:
Sorry about that -- I have made some installation changes for v2.0.0, but my intent was to *avoid* breaking the existing methods. I believe I've fixed the issue now, though, so the old `go get` method should be working again.

plaindocs:
Hahah, @jdkato that is even worse now I've update the build! ;-) Especially when your docs don't reference the `go get` as an official method.

jdkato:
It isn't an "official method" and hasn't been for almost a year (I changed this shortly after v1.0.0 was released) because of the inability to install a certain version (it's always simply `master` with `go get`), but I still try to avoid breaking existing installations when I make changes.

Again, sorry for the inconvenience.

PR #1130, opened 2019-11-05
Remove system messages from the atom feeds

ericholscher:
This stops us from publishing silly messages from RST

### DIFF
--- a/docs/_ext/atom_absolute.py
+++ b/docs/_ext/atom_absolute.py
@@ -56,6 +56,10 @@ def _rewrite_feed(app, tag):
     # Convert the content nodes to use absolute links
     for content in root.xpath('//atom:entry/atom:content', namespaces=nsmap):
         html_content = html.fromstring(content.text)
+        # Remove janky system messages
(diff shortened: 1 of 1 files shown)


plaindocs:
Unless I'm missing something, that doesn't work, because all of the content is escaped. At least look at the atom file locally shows system messages still in there.

ericholscher:
Are you looking at the right file? It's on the `_absolute.xml` that is processed. It doesn't run by default unless we're on prod, so you need to hack it a bit which is probably silly.

plaindocs (approved):
OKI doki, confirmed that the `system-message` isn't in https://deploy-preview-1130--dentist-canary-64055.netlify.com/blog/archive/tag/portland-2020/atom_absolute.xml

It *is* in a ton of other atom files though:

```
$ ack system-message _build/html/blog/archive/tag/*/atom.xml | wc
    200    1221   44346
```

So let's merge this and figure out a solution for the rest as well?

ericholscher:
Yea, we can only send email with the `absolute` ones. I could fix all the atom files, but _shrug_.

PR #1135, opened 2019-11-27
Debug Futures

plaindocs:


plaindocs:
:man_shrugging:

mxsasha:
Do we actually need this though? From reading https://python-future.org/_modules/future/standard_library.html I don't see any imports that this affects. Maybe it was there for historic reasons?

plaindocs:
I did wonder that. Futures is the Python 2 > 3 migration library right? And now we're running on 3 exclusively it should be safe to remove? 

Do you want to remove and push to this branch and see where we get?

plaindocs:
Alternatatively merge this so we can update stuff and make another PR to remove future.

PR #1139, opened 2019-12-02
Step back to Python 3.5 to keep Netlify happy?

plaindocs:
Netlify changed something, maybe around Select Build Image (beta) and builds were failing with Python 3.6.

This looks like it should work.


plaindocs:
FAO @bethaitman

mxsasha (approved):
Or upgrade to 3.7?
Our Python use is very light, so I don't have significant concerns with using 3.5 either.

plaindocs:
Yeah, I thought about switching to 3.7 but that would also require switching to the new build image on Netlify and I wanted a quick fix.

Issue #1085, opened 2019-09-15
New structure for the guide


bethaitman:
This is a proposal for a new way of organising all the content we have on the site: talks, guide entries, newsletter articles, and anything else that comes up. We'll never get this perfectly right, but I'd like to get a reasonably flexible structure that can cope with change and expansion.

I've tried to make it as task-based as possible, in the hope that this will map up to things that people are looking for, but I've not been able to do it totally.

All the headings are temporary: I don't think I have the names right yet.

- **Micro-level writing**: This means the nitty-gritty of writing, and includes things like:
  - grammar
  - plain english
  - very specific writing questions (one word vs another, using abbreviations, etc)
  - storytelling?
  - code samples?
  - other media - videos, diagrams, screenshots?
- **Page writing**: Writing at a higher level, roughly thinking about a "page". Includes things like:
  - task-based writing
  - specific doc types (how-tos, reference/API docs, tutorials, etc) and how to write them
- **Doc set writing**: Working at the level of thinking about a whole (or part of) a set of docs. Includes things like:
  - information architecture
  - refactoring and restructuring (legacy) docs
  - maintenance
- **Jobs and careers**
  - hiring
  - getting hired
  - related roles
  - career growth
- **Design and UX**
  - accessiblity
  - design
  - UI writing
- **Doc sites and tools**
  - docs-as-code
  - site design, metadata
  - DITA
  - automation and automated testing
  - Git
  - markdown and other formats
- **Analytics and metrics**
  - user research
  - gathering and acting on user feedback
  - measuring docs success
  - metrics
- **Interpersonal stuff**. _I think this label doesn't actually work - it won't be what people are looking for_
  - building a docs culture
  - building a community
  - working with writers (review, editing, teaching)
  - customer support, working with support
  - helping engineers to write
  - working with other roles (product managers, higher-ups, etc)
- **Planning**
  - prioiritisation
  - agile workflows
  - strategy
  - defining success

bethaitman:
Relevant links:

- [newsletter index](https://www.writethedocs.org/newsletter/#index-of-newsletter-topics)
- [video index](https://docs.google.com/document/d/1Ukb4XJNOzyaKeDOC4gpvuj9iYriqaNZunj9FOzs_j9Y/edit#)
- [the guide](https://www.writethedocs.org/guide/)

plaindocs:
What about Interpersonal Stuff -> Culture & Community

bethaitman:
Notes from the conversation:

[...]

PR #1162, opened 2020-02-03
Can we use sphinx indexes as a Categorization tool

plaindocs:
Some work on https://github.com/writethedocs/www/issues/1085#issuecomment-581578567

Short answer, not without custom development or a different directive.

image

Not showing titles is a deal breaker here.

* I didn't bother rewriting the video files because they get clobbered on update.
* I didn't add external links

@mxsasha @bethaitman


mxsasha:
I agree that this is useless. Maybe we can roll our own thing, I'll try to look into how complex that would be.

PR #1163, opened 2020-02-04
Fix unordered list formatting

kourtneyshort:
Fix unordered list formatting by adding a blank line before the list


plaindocs (approved):
Nice, thanks!

PR #1171, opened 2020-02-12
Add new logos and add year to site name

ericholscher:
This makes it easier to see which year is which,
hopefully cutting down on people thinking we're already sold out.


plaindocs (approved):
Seems great.

We could probably tweak the ticket page copy as well, but that shouldn't be a blocker.

PR #1181, opened 2020-02-21
Update pretalx2wtd to retrieve speaker info and other cleanup.

mxsasha:
The script now does its own API calls, and has a bit more
logging. Twitter and website are added to the sessions file
for Portland 2020. Order has changed.


mxsasha:
Nobody seems to have an avatar now, so we can add that feature once some speakers have uploaded one. Shouldn't be too hard now that we're retrieving full speaker info already for each speaker.

plaindocs:
This looks awesome. 

At least one speaker does have an Avatar, Lucy Cary.
They are also commented out on the speaker page till we have more.

mxsasha:
Oh, actually it's already in there. Do we want to make our own download, or just defer to pretalx? I don't like to manage automatic downloads, so I'm leaning to just do it like this:

```
  speakers:
  - name: Lucy Carey
    slug: lucy-carey
    avatar: https://pretalx.com/media/avatars/Screenshot_2020-02-19_at_09.56.26_pUIrHqc.png
    twitter: lucyrushi
```

plaindocs:
I'd be tempted to run it manually and update the git repo with the images so we're not dependent on Pretalx at build time. 

So wget the image and save to > images/slug.img 

What do you think @mxsasha

mxsasha:
> I'd be tempted to run it manually and update the git repo with the images so we're not dependent on Pretalx at build time.

I'll add that.

mxsasha:
I've added downloading of speaker images. Note that the location is derived from the slug in the speaker_photo template filter, so we don't actually need the avatar location in the sessions YAML, so I removed that. The speaker_photo template filter should be compatible with this.

mxsasha:
@plaindocs I think this is now ready for merging, can you approve the PR if you agree?

docs/_scripts/pretalx2wtd.py
```diff
@@ -1,55 +1,110 @@
...
-        yaml_obj.Representer = RoundTripRepresenter
-        yaml_obj.default_flow_style = False
-        with open(yaml_filename, 'w+') as yaml_file:
-            yaml_obj.dump(yamldoc, yaml_file)
+def convert_to_yaml(year, series, yaml_output, pretalx_slug):
+    if not os.environ.get('PRETALX_TOKEN'):
+        print('Error: PRETALX_TOKEN not found in environment variables.')
+        return
+    http_headers = {'Authorization': 'Token ' + os.environ['PRETALX_TOKEN']}
+    submissions_url = f'https://pretalx.com/api/events/{pretalx_slug}/submissions/?state=confirmed'
```
plaindocs, line 32: `{pretalx_slug}` needs to available here. Not sure if you use any of the other

mxsasha: I am confused. It is available. The script runs fine?

mxsasha:
@plaindocs I added the current output for the Portland 2020 sessions in this PR too. The script seems to run fine, and I think it was all done.

So I think we can merge this, which will also update the Portland 2020 sessions - you should probably have a look over https://github.com/writethedocs/www/blob/pretalx-improvements/docs/_data/portland-2020-sessions.yaml to see if it seems correct.

plaindocs (approved):
This is awesome!

PR #1182, opened 2020-02-21
Validate that all sessions are scheduled if flaghasscheduled is set.

mxsasha:
Also makes the errors in scheduling explicit build aborts,
rather than warning logs, as they should always be fixed.


plaindocs (approved):
Much better! <3

PR #1191, opened 2020-03-04
Call For Volunteers

alexfornuto:
## Important Change
Puts out the call for volunteers for Portland 2020

## Minor Change
- Updates to project README
- Clarifications for volunteer guidelines


mxsasha:
@alexfornuto is this still current? Looks like not updated for virtual conference, don't know if something else happened instead.

alexfornuto:
No, this is not current.

PR #1213, opened 2020-03-25
Update meetup team

mjang:
As both Catherine Heath and Deborah Barnard have declined the offer, I've removed the name for EMEA.


plaindocs (approved):
:+1: :-(

PR #1225, opened 2020-04-28
Add social rule about TWs

janine-c:
Also removed mention of monthly message because it's not a thing anymore

### DIFF
--- a/docs/slack.rst
+++ b/docs/slack.rst
@@ -69,13 +69,14 @@ Guidelines
 Our goal with Write the Docs is to be **the most welcoming community on the internet**.
 People who are new to our community in particular should be assisted and welcomed.
 
-Our :doc:`/code-of-conduct`, applies to all spaces associated with Write the Docs, which includes our Slack.
+Our :doc:`/code-of-conduct` applies to all spaces associated with Write the Docs, which includes our Slack.
 In addition, we have a few Slack-specific guidelines:
 
 Social rules
 ~~~~~~~~~~~~
 
 - **Use inclusive language** - Our community is multicultural, multiracial, multigendered, and generally diverse. When posting in channels, avoid using words such as "guys" to address a mixed group. Possible alternatives: "folks", "everyone", "documentarians".
+- **We're all documentarians** - The Write the Docs community includes developers, testers, PMs, DevRel folks, and many more. Remember that our Slack consists of more than just technical writers.
 - **Only participate when you can be constructive** - Try to be constructive in all that you do, and otherwise consider not participating in a conversation.
 - **Say hi** - The world is a better place with a flood of hello's.
 - **Don't be snarky** - We are all trying the best we can.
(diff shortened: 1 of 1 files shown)


docs/slack.rst
```diff
@@ -69,13 +69,14 @@ Guidelines
...
 
-Our :doc:`/code-of-conduct`, applies to all spaces associated with Write the Docs, which includes our Slack.
+Our :doc:`/code-of-conduct` applies to all spaces associated with Write the Docs, which includes our Slack.
 In addition, we have a few Slack-specific guidelines:
 
 Social rules
 ~~~~~~~~~~~~
 
 - **Use inclusive language** - Our community is multicultural, multiracial, multigendered, and generally diverse. When posting in channels, avoid using words such as "guys" to address a mixed group. Possible alternatives: "folks", "everyone", "documentarians".
+- **We're all documentarians** - The Write the Docs community includes developers, testers, PMs, DevRel folks, and many more. Remember that our Slack consists of more than just technical writers.
```
mjang: I love the spirit of the statement, but the list doesn't feel parallel. Maybe something like this, but I'm feeling like this could be improved too:

```suggestion
- **We're all documentarians** - The Write the Docs community includes developers, testers, program/product managers, developer relations specialists, and more. Remember that our Slack consists of more than just technical writers.
```

plaindocs: What about 

> **We're all documentarians** - The Write the Docs community is much more than a space for technical writers, we welcome developers, testers, PMs, DevRel folks, and many more.

And then if we want to add a specific mention of Slack, something along the lines of "think about folks in various roles might react to your comment about writing or programming"?

nbkaplan: I know why we're tempted to emphasize "not just tech writers," but it feels like we're scolding people. What if we include tech writers as just one of the examples, and not called out specifically: "The Write the Docs community welcomes everyone who wants to learn more about creating documentation. This includes developers, testers, customer support agents, technical writers, program/product managers, developer relations specialists, and more."

janine-c: How about this?

The Write the Docs community is much more than a space for technical writers. We welcome everyone who wants to learn more about creating documentation, including developers, testers, customer support agents, technical writers, program/product managers, developer relations specialists, and more.

I think it's a good idea to call out tech writers because I still see references to it being "that tech writer community."

plaindocs: Sounds great to me. And I think this whole thing came from a need to emphasize "not just tech writers" although no guarantee the writing it here will help.

I would nit pick

> technical writers, program managers, product managers, developer relations specialists, and more.

though

ddbeck: After reading this conversation, I wondered if there was a way _everyone_ could feel like they were being ~scolded~ included and I came up with: `The Write the Docs community is much more than any one job title.` But I'm fine with it either way, at least with the aforementioned nitpick. 😄

ddbeck (approved):

Bradamant3:
I do like Daniel's last suggestion, but not going to stand on making the change. And it's approved, soooo ...

ericholscher (approved):
This looks 💯. I'm 👍 on shipping.

ericholscher:
Going to merge this since there hasn't been any additional activity, and folks seemed in agreement.

PR #1230, opened 2020-05-04
Perhaps my date format was wrong?

bethaitman:
I can't see the newsletter on the blogs - did I mess up the date format?


ericholscher:
@bethaitman I see someone built the docs, they likely didn't publish because the date hadn't happened yet. I pushed the newsletter to publish at 5pm CET, so it should go out here soon.

ericholscher:
Hrm, that didn't work. I've set it back to 11am and hopefully it sends tomorrow.

bethaitman:
Rather odd. Hopefully will come through tomorrow!

PR #1236, opened 2020-05-07
Add emoji to the topics page

ericholscher:
This makes it a bit easier to tell what the content type of each link is

Preview here: https://writethedocs-www--1236.org.readthedocs.build/topics/#writing-words

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -44,6 +44,7 @@ extensions = [
     'ablog',
     'sphinxcontrib.datatemplates',
     'notfound.extension',
+    'sphinxemoji.sphinxemoji',
     'recommonmark',
 ]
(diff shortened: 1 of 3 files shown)


mxsasha (approved):
💯

plaindocs (approved):
:100:

bethaitman (approved):
You superstar, thank you for doing this! Huge improvement

plaindocs:
The other thing that might be interesting for this kinda content - is looking at a two column layout.

ericholscher:
@plaindocs Yea, I think there's a lot more we could be doing with this content. I'd like to get it properly tagged so that we can do a bit more with it on the detail pages as well. Baby steps tho :)

PR #1240, opened 2020-05-14
Update moderator list to everyone in the slack-mods channel

ericholscher:
We should keep this list and the membership of that channel consistent,
so that users know who they can reach out to.


ddbeck (approved):
(Approving for myself. I figure everyone named in the diff should have to approve before merging.)

janine-c (approved):
Looks great! Thanks, Eric. :)

plaindocs (approved):
:+1:

PR #1249, opened 2020-06-02
Add statement in support of black community 

ericholscher:
This is a mostly pre-approved draft for our website on the statement in support of the black community. Hopefully it will go out today.


mxsasha:
I don't see the Sphinx issue that pyup raises as affecting us, so I think we can go ahead with merging when ready regardless of it. Will look into it more soon.

Bradamant3 (approved):

PR #1288, opened 2020-07-31
Update welcome wagon doc

ericholscher:
This is converted from Gdocs, so probably needs some cleanup


ericholscher:
Gonna merge for the email, and we can cleanup as we go. I think I got most everything.

PR #1304, opened 2020-08-08
Clean up CI

plaindocs:
So something is super broke on the Travis CI. Our tox environments look like they aren't being run at all, so none of the checks are actually happening. I'm doing a mix of investigation and fixing problems where I find them here.

Looks like the magic that was turning the environment variables into tox environments and running them is just not working. 

https://travis-ci.org/github/writethedocs/www/jobs/716098168

So all of the builds are just running `tox` and reporting success, but doing nothing.

Yeah, the main issue was a python version mismatch,  3.5 in Travis, 3.6 in Tox, which means I broke this 8 months ago in https://github.com/writethedocs/www/pull/1139 ! :-D


plaindocs:
And of course we can't merge this without fixing t"HTML-Proofer found 82 failures!" because it'll break deployment.
Probably making the link checker also an allowed failure for now is the way to go.

docs/hiring-guide/index.rst
```diff
@@ -117,11 +117,15 @@ Develop your capabilities with these professional skills and tools.
 
 Skills and Knowledge:
 
+.. vale off
```
ericholscher, line 120: Huh, cool.

ericholscher (approved):
Lots of good changes in here, we should merge it before it gets too stale and has a ton of conflicts.

plaindocs:
Yeap, thanks for the ping! I'll merge now, and we should revisit ac9858b later

PR #1306, opened 2020-08-08
Aug2020 book club updates

djwfyi:
Adding in the current _Don't Make Me Think_ title, upcoming titles, and reformatting the rest of the page since we are done with previous timelines.


plaindocs:
This is great! Would you mind updating the PR so we're not deleting the 3 videos files? I think there is probably a gltich in the build process somewhere.

PR #1307, opened 2020-08-09
Adding back deleted video files

djwfyi:


djwfyi:
I guess this one's not needed.

PR #1313, opened 2020-08-11
Use people instead of tribe, and add a linter

plaindocs:
I'm open to other terms if folks have better suggestions, and I opted to edit the direct quote also, as it is central to "our" story.


mxsasha (approved):
💯

PR #1314, opened 2020-08-13
Fixed spacing problem

ctauchen:
In two places there was no space between a word with strong emphasis
and the next word. Looks like an escape character
was affecting the output.

Ex. **bold**\ nextword ==> **bold** nextword


plaindocs (approved):
Thanks! :-)

PR #1315, opened 2020-08-17
Update staff pics Portland 2020

thatdocslady:


plaindocs (approved):
Would make sense to have them all lower case I think.

mxsasha:
Fixing something, don't merge yet

mxsasha:
I trimmed down the image sizes a bit, 4000x4000 images of 4 MB seemed a bit excessive - and this will be part of our git repo forever :)

plaindocs:
We should set up https://github.com/marketplace/actions/image-optimizer  or something probably.

Issue #1316, opened 2020-08-17
Unable to build with specified versions of ablog, yamale


mjang:
# Problem:

The website build instructions in the current README fails due to version issues on the following packages in requirements.txt:

- ablog
- yamale

## How to replicate:

When doing a "local build" of the website, per https://github.com/writethedocs/www

I generate a virtual environment per 

```
virtualenv --python=/usr/bin/python3.5 venv
```

I activate the virtual environment on Linux (Ubuntu 16.04)

```
source venv/bin/activate
```

I try installing Sphinx:

```
pip install -r requirements.txt,
```

I get the following error:

```
ERROR: Could not find a version that satisfies the requirement ablog==0.10.6 (from -r requirements.txt (line 3)) (from versions: 0.1, 0.2, 0.2.1, 0.2.2, 0.2.3, 0.3, 0.3.1, 0.4, 0.5, 0.5.1, 0.6, 0.6.1, 0.6.2, 0.6.3, 0.6.4, 0.6.5, 0.7.0, 0.7.1, 0.7.2, 0.7.3, 0.7.5, 0.7.6, 0.7.7, 0.7.8, 0.7.9, 0.7.10, 0.7.11, 0.7.12, 0.8.0, 0.8.1, 0.8.2, 0.8.3, 0.8.4, 0.9.0, 0.9.1, 0.9.2, 0.9.3, 0.9.4, 0.9.5)
ERROR: No matching distribution found for ablog==0.10.6 (from -r requirements.txt (line 3))
```

## Workaround:

I try working around the error by specifying `ablog==0.9.5` in requirements.txt, I try installing the requirements again, and I get this error:

```
ERROR: Could not find a version that satisfies the requirement yamale==3.0.1 (from -r requirements.txt (line 14)) (from versions: 1.2.0.macosx-10.10-intel, 1.3.0.macosx-10.10-intel, 1.3.1.macosx-10.10-intel, 1.1.3, 1.2.0, 1.2.1, 1.3.0, 1.3.1, 1.4.0, 1.4.1, 1.5.0, 1.5.2, 1.5.3, 1.5.4, 1.5.5, 1.5.6, 1.6.0, 1.6.1, 1.6.2, 1.6.3, 1.6.4, 1.7.0, 1.7.1, 1.8.0, 1.8.1, 1.9.0, 1.10.0, 1.10.1, 2.0, 2.0.1, 2.1.0, 2.2.0, 3.0.0)
[...]

PR #1318, opened 2020-08-17
Update python version in README

plaindocs:
@mjang that bug report is beautifully clear.

Turns out that because I spent some time looking at #1304 I figured out this one pretty quickly.

Fix for #1316


mjang:
Fix verified! Approving.

mjang (approved):

PR #1363, opened 2020-10-18
Fix unconf start time

ericholscher:


mxsasha (approved):
Changed in hopin too

ericholscher:
@mxsasha ha, I was going to change it in hopin, and it magically changed out when I went to edit it :D

PR #1375, opened 2020-11-08
Fix formatting for speaker abstract

bravegrape:
Adds HTML formatting so that the lists in the abstract are rendered correctly.

Before:

image

After:

image


plaindocs (approved):
Danke! :-)

PR #1378, opened 2020-11-10
Restructure Bias section and full copy-edit.

jaredmorgs:
I added in the reference to the Atlassian Design resource and broke out the references in this section into guides and resources (because there was a mixture of types, and I found it a PITA to use the first time I read this page).

I also noticed some innocent typos on the page so ran it through a grammar linter and cleaned up the language.


plaindocs (approved):
Solid edits, thanks @jaredmorgs .

I quite like frequenty though. Fifty, sixty, oh I can't be bothered, frequenty.

PR #1382, opened 2020-11-18
Add a 2021 Portland landing page.

ericholscher:
This:

* Creates baseline CSS & HTML for the 2021 sites
* Updates the colors to Purple on the landing page
* Updates the config to have new variables
* Adapts a few templates to be more resilient to changes

TODO: 

* I adapted the SVG icons a bit, but didn't finish them. We should probably clean them up before we ship it. We can probably just do complimentary colors like we did previously.


mxsasha:
The whole 2021 _templates directory is identical except for the background of the hero element. Perhaps we could keep using the 2020 templates if we extract that gradient colour?

ericholscher:
Yea, there's definitely some cleanup to do there. It shouldn't be hard to template that out and delete all the 2017-2021 into a single directory.

PR #1403, opened 2020-12-16
Ship the 2021 full Portland site

ericholscher:
This does a bit of cleanup on the schedule and config,
and a few other tweaks.
I think things are looking good here tho.

TODO:

* We should make the team page generated from YAML
* The menus should be a bit more dynamic based on flags
* Update speaking tips page for online events

Otherwise, it feels like our data-driven approach is working nicely.


ericholscher:
Gonna ship this so it's ready when people hit the CFP.

PR #1412, opened 2021-01-08
Fix Mapbox link

rachfop:
Mapbox link not working the new link is the correct url


plaindocs (approved):
Thanks for fixing! <3

PR #1435, opened 2021-02-14
Migrate attendee resources for virtual confs from GDocs

thatdocslady:


docs/include/conf/virtual/talk-recording-guide.rst
```diff
@@ -0,0 +1,136 @@
...
+
+All talks are going to be processed and uploaded in advance, so it's especially critical that you complete your recording on time! Please email us if you have any questions.
+
+All visual and audio content (including videos, room background, and slides) must adhere to the Write the Docs Code of Conduct.
+
+Video requirements
+------------------
+
+- Duration: 30min (+- a few minutes is ok)
+- No cutting required if the video is in a single file
```
mxsasha: I am confused by what this means. Especially because later on, we say we can cut parts out of the video if needed:
> While you are recording the talk, if you run into a problem with your narration but you don’t want to stop recording, we can edit it later! Take a breath, repeat the last 2-3 sentences, and when you send us the video file please write down time markers of places where edits are needed.

mxsasha (approved):
I think this is great, with the one comment in-line.

docs/include/conf/virtual/talk-recording-guide.rst
thatdocslady: Hmm, well the point is that if they are recording a continous single file they don't need to trim the lead-in and lead-out footage (because we instruct them to wait 10-20 seconds before and after the talk). Feel free to suggest a different phrasing, maybe we can replace "cutting" with "editing"?

plaindocs:
Happy to let you folks merge this in. :+1:

docs/include/conf/virtual/talk-recording-guide.rst
thatdocslady: @mxsasha what do you think about my suggestion?

ericholscher: Perhaps:

```suggestion
- Single video file without edits
```

mxsasha: I like Eric's suggestion. The current text makes me think the opposite: that the file I deliver as a speaker must not require any further cutting.

PR #1447, opened 2021-02-26
Mentoring guidelines cleanup

erin-rtfm:
Adjusted from first person to third person. Cleaned up and reorganized suggestions.


plaindocs (approved):
You're a star thanks! This is much improved.

PR #1455, opened 2021-03-09
Set integrated schedule and lightning talk form links.

mxsasha:
Note that integrated schedule link does not show up yet,
it will after flaghasschedule is set.


plaindocs (approved):
LGTM

PR #1385, opened 2020-11-21
Clean up Tox config a little

plaindocs:
- Run sphinx only once 
- remove allowlist related bad config

PR #1463, opened 2021-03-17
Continuous integration hacking

mxsasha:
(See comment)


mxsasha:
Vale check currently broken due to https://github.com/errata-ai/vale/issues/340

plaindocs:
Do you think we should switch sphinx to strict mode? 

```
-W                turn warnings into errors
```

mxsasha:
> Do you think we should switch sphinx to strict mode?
> 
> ```
> -W                turn warnings into errors
> ```

Yes, that's part of the PR now :)

plaindocs:
You're a star! I'm excited to see how this turns out!

mxsasha:
I think this is now ready for review. Tox is no longer used, there are now four GitHub actions:
* One running YAML validation
* One running vale
* One running sphinx
* A check for internal links that runs after the sphinx build

You can see it in action in the checks on this PR. Run time is a few minutes, faster than the RTD build.

This PR also includes:
* Adding vale 2.9 support which means including more styles (none are shipped in vale anymore)
* Updating YAML schemas so that they validate
* Updating a fair number of broken internal links
* Fixing any remaining sphinx warnings

Sphinx is set to fail on warnings. Suggestion: merge soon, evaluate for a while, make required for merging later.

Separate idea is to periodic run external link checks against master.

Includes #1385, obsoletes #1386.

plaindocs (commented):
This is SWEEEEEEEET

- [x] What is the 5th vale action?
       image
- [x] Does html in n RST title get passed through? TIL
- [x] LOL

             tiger-like:  tigrine
             viper-like:  viperine
              vulture-like:  vulturine

Everything else looks great to me.

mxsasha:
> This is SWEEEEEEEET

🎉

>     * [ ]  What is the 5th vale action?

One is the actual job that starts vale, the second is the result of the report. This is how the official vale github runner does it, and I like to use that one because it'll leave in-line comments for style issues. So you don't have to read through the raw text output.

I would still like inline comments for Sphinx and links warnings/errors too, but that's for later.

>     * [ ]  Does html in n RST title get passed through? TIL

Don't know what you're talking about?

>     * [ ]  LOL

On my todo list is also to see if we can and should enable the full proselint set and fix the open issues with that. Some would actually be good, like avoiding "12 pm" (which we have in a few places, might be legacy).

plaindocs:
html in RST title is about having `<i>Sponsored</i> Talk title`  in `docs/videos/portland/2019/documentation-for-good-riona-macnamara.rst`

Got you on the Vale runner, makes sense, just looks odd now when there is no warnings. :+1:

plaindocs (approved):
Let's do this!

PR #1465, opened 2021-03-18
Hide the specific event times when we have a live schedule

ericholscher:
This will stop these from being out of sync with the main schedule.
We should only have this data in one place,
so let's not include it both places.


mxsasha (approved):

plaindocs (approved):
Sneaky, I like it.

plaindocs:
:+1:

PR #1477, opened 2021-04-05
Add Job Fair end time, WTD Portland 2021

mjang:


ericholscher (approved):
I adjusted it onto its own line, so it matches how it's done other places in the schedule 👍

plaindocs (approved):

PR #1506, opened 2021-04-25
Adding information about error messages

catherineheath:
@Bradamant3


catherineheath:
Has been added in a different place.

PR #1514, opened 2021-04-27
Update stats!!

plaindocs:
- [x] Apply to Australia


docs/conf/prague/2021/press-kit.rst
```diff
@@ -38,9 +38,8 @@ Useful numbers
 
 We have some stats about attendees, membership, and website hits over on the following blog posts:
 
-- :doc:`/blog/write-the-docs-2017-stats/`
-- :doc:`/blog/write-the-docs-2016-year-in-review/`
-
+- :doc:`/blog/2021-january-community-update/`
+- :doc:`/blog/write-the-docs-2020-stats/`
```
ericholscher: We should make this an include across all conferences :)

plaindocs: That does make sense.

ericholscher (approved):
Looks good. We should also apply it to Portland, no?

plaindocs:
Sure, I care more about next year than this year, but sure.

PR #1515, opened 2021-04-27
Use v2 for RTD yaml with right builder

ericholscher:


plaindocs:
Been there a while huh?

plaindocs (approved):
Preview works

PR #1517, opened 2021-04-28
Set postconf for Portland 2021

mxsasha:


plaindocs (approved):
Looks good to me.

I fixed the unrelated vale warning in #1518.

PR #1527, opened 2021-05-05
Remove broken RST links from guide index.rst

mitches-got-glitches:
It looks like these links have moved to "Introduction to RST/Resources" and that these were intended to be removed, but for whatever reason, they were not.


plaindocs (approved):
Sounds right to me. Thanks!

PR #1530, opened 2021-05-07
Fixed link by adding Vale config file extension

ctauchen:
I fixed a bad link to the Write the Docs [Vale configuration file](https://www.writethedocs.org/guide/tools/testing/#vale).

Bad link: https://github.com/writethedocs/www/blob/master/.vale
Fixed link: https://github.com/writethedocs/www/blob/master/.vale.ini

Thanks!


plaindocs (approved):
Thanks!

PR #1537, opened 2021-05-23
Add Portland 2021 sketchnotes and videos

mxsasha:
Flickr's URLs contain some random hashes, so we can't easily include a flickr ID in the config. Hence a new sketchnote URL and image in the sessions file.

https://writethedocs-www--1537.org.readthedocs.build/videos/portland/2021/writing-a-perfect-technical-tutorial-jessica-garson/ for an example

### DIFF
--- a/docs/_data/portland-2021-sessions.yaml
+++ b/docs/_data/portland-2021-sessions.yaml
@@ -3,6 +3,9 @@
   series: Write the Docs Portland
   series_slug: portland
   year: 2021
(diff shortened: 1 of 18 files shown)


ericholscher (approved):
Looks 💯  -- Thanks for getting these on the site. 

We should also tweet something about this when it goes out. Need a good process for that :)

plaindocs:
Awesome

PR #1538, opened 2021-05-24
Enable Prague 2021 tickets

mxsasha:
https://writethedocs-www--1538.org.readthedocs.build/conf/prague/2021/tickets/

The tito event is not live yet, so only visible if you're logged in to tito with admin permissions.


mxsasha:
Obsoleted by #1540

PR #1545, opened 2021-06-03
Survey results take 1

plaindocs:
- [x] needs images
- [x] CSS for `<div class="question docutils container">`

:arrow_forward: [Direct link to the 2020 results](https://writethedocs-www--1545.org.readthedocs.build/surveys/salary-survey/2020/) :arrow_backward:


docs/surveys/salary-survey/2020.rst
```diff
@@ -0,0 +1,1876 @@
...
+ “neutral”. In fact, only 1 respondent - from the “partially remote,
+partially onsite” segment - reported feeling “negative” about their work
+location, and no one reported feeling “very negative”.
+
+.. figure:: images/2020/feelings-location-unchanged.svg
+   :alt: Figure: Feelings about Work Location
+(where work location is unchanged since COVID-19)
+
+   Figure: Feelings about Work Location
+(where work location is unchanged since COVID-19)
```
plaindocs:
```suggestion
.. figure:: images/2020/feelings-location-unchanged.svg
   :alt: Figure: Feelings about Work Location (where work location is unchanged since COVID-19)

   Figure: Feelings about Work Location (where work location is unchanged since COVID-19)
```

thatdocslady:
I skimmed both 2019 and 2020 results pages and they look great (didn't read every single word but I assume this review is for look and feel). 

The only weirdness is that the Surveys page itself shows both the results and the survey links for both years, in a seemingly-random order. If there's a way to hide completed surveys and also order the survey years in latest-first order, it would be great.

thatdocslady (approved):
Approved with small comments. Also, maybe we can squash all those commits before merging just for repository hygiene purposes ;)

docs/_scripts/fix_gdocs2rst.py
```diff
@@ -0,0 +1,24 @@
+in_rubric = False
```
ericholscher, line 3:
```suggestion
# Script used to migrate gdocs into RST. 
# Historically used for the salary survey, but might be useful otherwise. 
in_rubric = False
```

docs/organizer-guide/index.rst
```diff
@@ -87,6 +87,43 @@ Information about contributing interview content for the community.
 
    community-spotlight-interviewing-guide
     
+Salary Surveys
```
ericholscher, line 90: This shouldn't go in the index. It should be a page on surveys.

ericholscher (commented):
Agreed w/ Mikey, we should do an explicit TOC in the surveys index. Will make the ordering more reasonable.

docs/organizer-guide/index.rst
plaindocs: Dammit people, don't point out when I cut corners. :-D 
Also, yup, will fix

ericholscher (approved):
Sorry, forgot the 👍 -- definitely only small nits, and we should ship it!

plaindocs:
@ericholscher what is the story with redirects? If redirecting the survey page to the results is possible we could just do that?

ericholscher:
We can do that, but it feels weird. We should just cross-link them I think?

plaindocs:
oki doki

plaindocs:
image

ericholscher:
Set to auto-merge 🎉

PR #1559, opened 2021-07-19
Extract Portland 2021 welcome wagon

mxsasha:
Extracted the Portland 2021 content and applied it for Prague 2021 too. It does mention social event now, so might need more finetuning later to use in other events.


docs/include/conf/virtual/welcome-wagon.rst
```diff
@@ -0,0 +1,362 @@
...
+
+- Select the **Video Camera** icon to turn off your video.
+- Select the **Monitor** icon to share your screen. Multiple people can share their screens at once.
+- To make someone’s screen larger, double-click on it.
+- Select the **Settings** icon to change your audio or video settings.
+
+Writing Day
+-----------
+
+`Writing Day <https://www.writethedocs.org/conf/{{ shortcode }}/{{ year }}/writing-day/>`__ takes place on Sunday, the day before the conference starts. You can join in anytime between 9:30am and 5pm {{tz}}.
```
ericholscher: These times should probably be templated.

docs/include/conf/virtual/welcome-wagon.rst
```diff
@@ -0,0 +1,362 @@
...
+   :local:
+   :depth: 1
+   :backlinks: none
+
+Where to get help
+-----------------
+
+Conference Intro and Tour
+~~~~~~~~~~~~~~~~~~~~~~~~~
+Join our our Conference Intro and Tour on Sunday, April 25th at 12:30 pm PST by going to **Sessions** and selecting **Conference Intro and Tour**. Your Welcome Wagon will take you on a tour of the virtual conference platform and show you how to navigate the conference. Learn tips and tricks, meet other attendees, and get to know how the conference works. This session is open to everyone, whether this is your first conference or your fifth.
```
ericholscher: These times should be templated too -- are we doing this for Prague as well? I don't know if we have in the past.

ericholscher (approved):
Solid refactor 👍

docs/include/conf/virtual/welcome-wagon.rst
mxsasha: I think virtual conference tour was new this year in Portland and a great success, so let's assume it will happen for now.

thatdocslady (approved):
Great to extract and reuse this! I have some feedback about the content but that's for the WW team to handle.

PR #1577, opened 2021-08-24
Prague 2021 schedule

plaindocs:
- [ ] TBC by speakers
- [x] Fix time bug


mxsasha (approved):
Looks good!

- Might want to change the "see the talks" button to "see the schedule" or whatever text we usually have?
- It looks a bit odd that two speakers have pronouns in their name, but the rest does not. Maybe no easy fix unless we make pronouns a proper part of our data structure.

mxsasha:
For time, seems we're expecting time as a string, i.e. `'10:25'`, not `10:25`.

plaindocs:
* From next conf we should add pronouns in Pretalx, and to the sessions data structure. :+1: 
* Good call on the button also.

docs/_data/prague-2021-schedule.yaml
```diff
@@ -0,0 +1,79 @@
+writing_day:
+- time: '10:00'
```
thatdocslady, line 2: This looks a bit later than we usually started writing day, but I'm not actually mad at it... Maybe we can trim the time between opening the platform and doing the introduction to 30min? We tried something like that at some point and it was good....

docs/_data/prague-2021-schedule.yaml
```diff
@@ -0,0 +1,79 @@
...
+  title: 'Platform opens'
+- time: '11:00'
+  title: Morning introduction - Introduction to Sprints & Project Introductions
+- time: '11:15'
+  title: Team up and start working
+- time: '14:30'
+  title: Lunch Break
+- time: '19:00'
```
thatdocslady: I think that's a really late hour to finish, maybe we can shorten the day a bit to end at 17:00?

Also, @ddbeck @bethaitman when do you want to do the WW tour session?

docs/_data/prague-2021-schedule.yaml
```diff
@@ -0,0 +1,79 @@
...
+  title: End of Writing Day
+
+talks_day1:
+- time: '9:00'
```
thatdocslady, line 15: @ddbeck @bethaitman when do you want to do the second WW tour? In Portland we had it at :15 past the opening hour and it ran for about 30min

docs/_data/prague-2021-schedule.yaml
```diff
@@ -0,0 +1,79 @@
...
+  title: Platform opens
+- time: '10:00'
+  title: Write the Docs Team - Introduction to Write the Docs
+- time: '10:45'
+  title: <b>Unconference Starts</b>
+- time: '10:45'
+  slug: hitchhiker-s-guide-to-documentation-tools-and-processes-lukas-reu-ner
+- time: '11:25'
```
thatdocslady: Is this time slot shorter intentionally? Feels like it would hurt the Q&A time, unless you know that the talk itself is short....

docs/_data/prague-2021-schedule.yaml
```diff
@@ -0,0 +1,79 @@
...
+- time: '13:00'
+  title: Lunch (60 mins)
+- time: '14:00'
+  title: Lightning talks (30 mins)
+- time: '14:35'
+  slug: when-documenting-is-designing-how-to-assist-api-design-as-a-technical-writer-fabrizio-ferri-benedetti
+- time: '15:20'
+  slug: adventures-in-setting-up-a-knowledge-system-for-a-research-group-victor-sluiter
+- time: '16:00'
+  title: Break (10 mins)
```
thatdocslady, line 46: I see what you're doing <3 I'm still not a fan of cutting talk slots shorter, and my dyslexic eyes would be really confused by all the odd time stamps... what do we think about keeping the time slots 45min and taking 15min breaks? yes it'll push the schedule forward by a little bit but I feel like it might be easier to navigate. We can also cut lunch to 45 min if we want to catch up, since most people won't go out to eat...

thatdocslady (commented):
Looks good so far, a few nits and a somewhat-important decision to make about breaks

docs/_data/prague-2021-schedule.yaml
plaindocs: I'm against cutting lunch, I think giving people the "ok" to step away from the monitor is valuable.

docs/_data/prague-2021-schedule.yaml
```diff
@@ -0,0 +1,78 @@
+writing_day:
```
thatdocslady, line 1: The CEST is aligned weirdly under the starting time, not sure if it's actually needed inline...

thatdocslady:
The schedule itself looks great technically, now we just need to decide whether to explicitly include the 5min breaks. I'm leaning towards YES because it'll help break down the time slots visually. So we'll end up with 40min talk slots (with implied Q&A) and 5min breaks as a standard, plus the longer breaks and lunch. How does that sound?

plaindocs:
Yeah, I'll let @mxsasha handle the inline time zone. And I'm planning to add the breaks in today.

docs/_data/prague-2021-schedule.yaml
```diff
@@ -0,0 +1,96 @@
...
+  title: Lunch Break (60 mins)
+- time: '17:00'
+  title: End of Writing Day
+
+talks_day1:
+- time: '9:00'
+  duration: '1:00'
+  title: Platform opens
+- duration: '0:45'
+  title: Write the Docs Team - Introduction to Write the Docs
```
thatdocslady, line 19: Missing a 5min break for consistency

docs/_data/prague-2021-schedule.yaml
```diff
@@ -0,0 +1,96 @@
...
+- title: <b>Unconference Starts</b>
+- duration: '0:40'
+  slug: hitchhiker-s-guide-to-documentation-tools-and-processes-lukas-reu-ner
+- duration: '0:10'
+  title: Break (10 mins)
+- duration: '0:40'
+  slug: more-than-words-reviewing-and-updating-your-information-architecture-lana-brindley-she-her
+- duration: '0:5'
+  title: Switch speakers (5 mins)
```
thatdocslady: I thought we were just going to use "Break" for all of them? Not terribly emotional about it ;)

docs/_data/prague-2021-schedule.yaml
```diff
@@ -0,0 +1,96 @@
...
+- duration: '0:40'
+  slug: customer-feedback-is-the-fuel-in-our-engine-kevin-hwang
+- duration: '0:25'
+  title: Day 1 wraps up
+- time: '18:00'
+  duration: '2:00'
+  title: <b>Social Event</b>
+- title: <b>Social Event Ends</b>
+
```
thatdocslady, line 59: I can't figure out how to comment on a different part of the file (I blame Fridays) but there's a missing bit in the location of the Social Event after the Monday schedule. we can just write "Spatial Chat, details will be posted during the conference" or something generic like that :)

PR #1580, opened 2021-08-26
Update Aus 2021 website for cancellation

mxsasha:


thatdocslady:
Looks good, just a couple of comments:

- Would it be better to revert the site back to the landing page format and build the cancellation notice on top of that? Even though the front page is updated, there's still access to all the other pages in the site, which might be confusing. Not sure if that's easy to do technically, just offering an alternative...
- There's a missing gap between the cancellation post button and the black banner: 
Screenshot 2021-08-26 at 14 07 11

mxsasha:
>     * Would it be better to revert the site back to the landing page format and build the cancellation notice on top of that? Even though the front page is updated, there's still access to all the other pages in the site, which might be confusing. Not sure if that's easy to do technically, just offering an alternative...

Done, that's much better. Great suggestion!

>     * There's a missing gap between the cancellation post button and the black banner:
>       Screenshot 2021-08-26 at 14 07 11

Fixed by making it a landing page it seems.

thatdocslady (approved):
Excellent, looks much cleaner now! Should we wait until the mailer is sent or is it safe to merge today?

mxsasha:
I'd say merge now - the news post is already on the website anyways, and it's better if the website is updated before people get the mail.

PR #1582, opened 2021-08-30
Allow schedule YAML to have duration instead of start time

### LINKED ISSUE #1581: Allow schedule YAML to have duration instead of start time
The conference schedule YAML could support a duration instead of start time (except for the first event of the day), to make it easier to edit. This will make the current google sheets schedule obsolete, along with the spreadsheet2schedule script, and make the YAML the single source of truth that is also practical to edit.

Only downside: if you enter the duration in the YAML, you need to build the site to see the calculated start times of every item. If this is very impractical, we can add a script that does the calculation separately and prints the result.

We'd also need a per conference setting of 12h/24h time display.

/cc @plaindocs

mxsasha:


mxsasha:
Wrong base branch

PR #1585, opened 2021-09-06
2021 organizer changes

batamig:


plaindocs (approved):
Thanks @batamig !

Was the meetup renamed on meetup.com? Or is this a new one? If so does the old one need deleting?  Cheers

batamig:
> Thanks @batamig !
> 
> Was the meetup renamed on meetup.com? Or is this a new one? If so does the old one need deleting? Cheers

yep we renamed it on meetup.com. we should probably rename this file itself as well, it's less critical and i wasn't sure how to do that, and where it might be referenced. thanks for merging!

PR #1588, opened 2021-09-07
Add CfP review criteria

mjang:
The Linux Foundation publishes their proposal review guidelines. I think it's in the spirit of open source.

I think we should consider doing something similar, and have drafted a proposed PR.

I've left certain dates blank (see lines 16-20). I'm not even certain that we should have deadlines for each of these items, individually. But I think it's worth our consideration.

The Linux Foundation apparently has a standard "review period" of [3 weeks](https://events.linuxfoundation.org/kubecon-cloudnativecon-north-america/program/scoring-guidelines/#overview).

Their "50%" milestone is 10 days into that period. Personally, I like the idea, as IMO it's important to avoid leaving these reviews to "the last minute".  But I think it is more prescriptive than what we normally do.


plaindocs:
I just lost 45 mins editing this into another PR. :facepalm: Will have another bash tomorrow.

PR #1591, opened 2021-09-09
Lint 'this' in links

plaindocs:
Set as a warning only, it's pretty basic.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,6 @@ jobs:
         uses: errata-ai/vale-action@v1.3.0
         with:
           files: docs/blog/
-        with:
           onlyAnnotateModifiedLines: true
 
   yaml-validate:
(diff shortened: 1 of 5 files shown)


plaindocs:
@mxsasha I noodled on this, but don't really think it is mergeable as applies to the docs as a whole.

Do you have any thoughts on either restricting it to changed files only, or making it more specific so it is useful?

i guess looking for `[this post|link|page]` kinda thing would be more specific than just `this` in a link.

mxsasha:
Hmm, I'm not sure, I don't know Vale too well. I'd have to prod at it a bit. Running it only, or more strict, on files actually changed in a certain PR seems like an interesting option in general. Or at least exclude old conference sites where we're not going to fix anything anymore.

plaindocs:
We're using the Vale action right? 

>  files: __onlyModified: Lint only files that have been modified within a PR.

From https://github.com/errata-ai/vale-action#files-default-all, and I wonder if it might be worth duplicating the action, so that some check everything and some check just the modified files? Not much in it.

plaindocs:
I'm going to stop caring about this.

PR #1604, opened 2021-09-30
Add Writing Day project "Open Web Docs and MDN"

teoli2003:
cc/ @elchi3, @ddbeck
This is our (Open Web Docs) for the Prague 2021 Writing Days. 

We are preparing documents about practical activities that we will link to the 4 projects before Sunday, but I think we can already advertise this!


plaindocs (approved):
Nice one @teoli2003 !

PR #1608, opened 2021-10-05
Spelling take 3

plaindocs:
This is my third and maybe final attempt at spell checking our stuff.

[Codespell](https://github.com/codespell-project/codespell) looks for known errors, not unknown words, so has less false positives.

1. In the `venv`, install with `pip install codespell`
2. Run using `codespell --skip="*.svg,*.js,*.css,*.scss,docs/_data/201[5|6|7]*" --ignore-words=codespell/ignore.txt docs`

Presumable something like this for the action?

```yaml
  spellcheck:
    runs-on: ubuntu-latest

    steps:
    - uses: actions/checkout@v2
    - name: Set up Python 3.8
      uses: actions/setup-python@v2
      with:
        python-version: 3.8

    - name: Install dependencies
      run: |
        python -m pip install --upgrade pip
        pip install -U codespell

    - name: Check spelling
      run: codespell --skip="*.svg,*.js,*.css,*.scss,docs/_data/201[5|6|7]*" --ignore-words=codespell/ignore.txt docs
```

Can replace #1554

PR #1621, opened 2021-11-11
Use the codespell GH action

plaindocs:
As per a comment in #1608 this gives us annotations and coloured output.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,23 +43,11 @@ jobs:
       run: docs/_scripts/validate-yaml.sh
 
   spellcheck:
-      runs-on: ubuntu-latest
-
-      steps:
-      - uses: actions/checkout@v2
-      - name: Set up Python 3.8
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.8
-
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install -U codespell
-
-      - name: Check spelling
-        run: codespell --skip="*.svg,*.js,*.css,*.scss,docs/_data/201[5|6|7]*" --ignore-words=codespell/ignore.txt docs
-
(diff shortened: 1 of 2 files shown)


plaindocs:
hmmm, that didn't work

```
invalid value workflow reference: references to workflows must be rooted in '.github/workflows'
```

mxsasha:
Maybe the indentation is off? Looking at the diff.

plaindocs:
~I don't actually see a huge improvement~

image

plaindocs:
@peternewman is this the expected output?

mxsasha:
The annotation is visible in-line in Files Changed on https://github.com/writethedocs/www/pull/1621/files#diff-c22dc5bd8c63d1494b82f44acea240bc5f1016bf5657c0214ac1259964569863

plaindocs:
Seems unintuitive to me to need to specify an image to run it on when the action is already defined, but that is nothing to do with this convo. @mxsasha approve if you're happy and we'll merge it in.

peternewman:
You can also see all the annotations in the action runner:
https://github.com/writethedocs/www/actions/runs/1448663886

plaindocs:
Ooh, @peternewman thank you once again, that is indeed nice

mxsasha (approved):
Thanks, this is a nice improvement :)

peternewman:
> Seems unintuitive to me to need to specify an image to run it on when the action is already defined, but that is nothing to do with this convo. @mxsasha approve if you're happy and we'll merge it in.

It doesn't work with a default if not then? I've never really played and just copy-pasted working ones most of the time.

plaindocs:
Yeah, it's a [required key](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idruns-on). I've not dug into whether you can specify it at the work-flow definition level not the repo you're using it in.

PR #1632, opened 2021-12-08
Add a section on creating channels to Slack guide

ddbeck:
Does what it says on the tin, mostly. It also rearranges things a little, so that all the channel stuff sits together.

[Direct link to the built preview](https://writethedocs-www--1632.org.readthedocs.build/slack/)

### DIFF
--- a/docs/slack.rst
+++ b/docs/slack.rst
@@ -37,8 +37,13 @@ That doesn't mean everything gets lost, though - you can always react to a messa
 ``:suggest-for-newsletter:`` emoji to nominate it for inclusion in our next monthly newsletter. 
 Read the newsletter archives at :doc:`/newsletter` and subscribe if you want them directly in your inbox! 
 
+Channel guide
+-------------
+
+There are lots of channels. Here's an overview of interesting channels and how to find more.
+
 Default channels
-----------------
+^^^^^^^^^^^^^^^^
 
 Everyone who joins the Slack will be added to these channels:
 
@@ -50,7 +55,7 @@ Everyone who joins the Slack will be added to these channels:
 * **#intros** - Introduce yourself! Let people know you're here, and why you care about docs :)
 
 Other useful topical channels
------------------------------
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 * **#career-advice** - Should you take that job? Am I being paid enough? Ask here!
 * **#markup-the-docs** - Talk about markup languages. Markdown, rST, Asciidoc, etc.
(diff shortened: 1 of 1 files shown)


ddbeck:
@plaindocs do you have the power to add @ravindk89 as someone who can review PRs?

plaindocs:
@ddbeck yes, but either it takes time to propagate, or write access isn't sufficient.

ravindk89:
Looks like I _can_ review, I'm just not on the Reviewers list. I think.

docs/slack.rst
```diff
@@ -94,21 +70,65 @@ We have some helpful Slackbot responses that you can summon in any channel, at a
...
+
+If you didn't see a channel that matches your interests, make sure to browse the complete channel list in Slack.
+
+If you still can't find a suitable channel, you can propose a new channel. To propose a channel:
+
+1. Find two other people to cosponsor your channel idea. Ask in related channels to find people who share your interest.
+2. Together, come up with a channel name and write a channel description.
+3. Post in **#meta** requesting a channel. Be sure to share your proposed channel name, description, and to tag your cosponsors.
+
+The Write the Docs moderation team will review your proposal.
```
ravindk89, line 132: So open question - is it worth monitoring new channels for activity? e.g. if the channel starts and then goes quiet immediately thereafter. I know it can take time to build up momentum, but I am also _guessing_ we want to keep the number of channels somewhat manageable from a moderation perspective.

This is to say, should we explicitly include some guidance stating that WtD may archive channels which are idle for longer than 90 days, and to re-activate see the guidelines for creating a new channel.

plaindocs: In theory sounds great, but then you gotta a) monitor, and b) raise the issue of whether to close it...

ravindk89: Ah the ever present monitoring issue :) Closing out.

ravindk89 (approved):

docs/slack.rst
```diff
@@ -94,21 +70,65 @@ We have some helpful Slackbot responses that you can summon in any channel, at a
 
   - Returns the answer for one of our most frequently asked questions: why you can't access the full chat history in the Write the Docs Slack.
 
+Channel guide
+-------------
+
+There are lots of channels. Here's an overview of interesting channels and how to find more.
```
plaindocs: Nit pick, sorry.

How about splitting out "these are the default channels everyone is added to", and then "here are some interesting / topical channels"

plaindocs (approved):
Looks good, with a nitpick

docs/slack.rst
ddbeck: I'm not sure I understand the suggestion here. Do you mean for me to go from this outline:

- Channel guide
  - Default channels
  - Other useful topical channels
  - Channels that serve specific communities & their allies
  - Location-based channels
  - Creating channels

To this one, which promotes the defaults outside the channel guide section?

- Default channels
- Channel guide
  - Other useful topical channels
  - Channels that serve specific communities & their allies
  - Location-based channels
  - Creating channels

Or something else?

plaindocs: Sorry @ddbeck I'll be more clear. I'm super happy with your structure, I just find the mention of "interesting ones" just before the list of defaults, even in the overview, a bit jarring.

Your new structure of separating out the default ones I think is even better though.

And let me apologize again for the nitpick

ddbeck: @plaindocs I tried my new proposed outline and was sort of disappointed in it (it also seems jarring to have default channels _not_ in the channel guide). I put it back, then rewrote the intro. What do you think?

plaindocs: Love it. :shipit:

ddbeck:
OK, let's go with it so I can reply in #meta about this.

@janine-c if you have anything to add, please review anyway and I can do a follow up PR based on your feedback.

Thanks everyone!

PR #1634, opened 2021-12-13
Added a meetup 

Dule-martins:
Added a Writethedocs Nigeria meetup to the list of meetups.


mxsasha:
This file is actually already in the repo, so I'm closing this PR, but the meetup was not listed on the website because we did not have a section yet for Africa meetups. I added that in #1636

PR #1650, opened 2022-01-17
Add tips for preserving Slack conversations

ddbeck:


docs/slack.rst
```diff
@@ -27,16 +27,24 @@ Feel free to send a Pull Request to update `this page`_, if you want to include
...
-We don't pay for our Slack, which means that the community has access to only the most
-recent 10,000 messages. Paying for Slack is prohibitively expensive for a community
-like ours, but many people seem to feel a freedom in knowing that their words will disappear shortly.
+We don't pay for our Slack, which means that the community has access to only the most recent 10,000 messages.
+Paying for Slack is prohibitively expensive for a community like ours,
+but many people seem to feel a freedom in knowing that their words will disappear shortly.
 We understand that not everyone loves this decision,
-but it's the way things are currently set up and we don't plan to change it in the near future. 
+but it's the way things are currently set up
+and we don't plan to change it in the near future.
```
janine-c: Maybe we could also change "near future" to "foreseeable future" to make it a little clearer that fetch isn't going to happen?

ddbeck: Good idea! I tightened it up even more. "We understand that not everyone loves this decision,
but we don't plan to change it in the foreseeable future."

janine-c: Love it!

janine-c (approved):
Looks great! Thanks for doing this 😄

docs/slack.rst
ravindk89: We might even just drop the timing entirely.

docs/slack.rst
```diff
@@ -27,16 +27,23 @@ Feel free to send a Pull Request to update `this page`_, if you want to include
...
+Paying for Slack is prohibitively expensive for a community like ours,
+but many people seem to feel a freedom in knowing that their words will disappear shortly.
 We understand that not everyone loves this decision,
-but it's the way things are currently set up and we don't plan to change it in the near future. 
+but we don't plan to change it in the foreseeable future.
 
-That doesn't mean everything gets lost, though - you can always react to a message with the 
-``:suggest-for-newsletter:`` emoji to nominate it for inclusion in our next monthly newsletter. 
-Read the newsletter archives at :doc:`/newsletter` and subscribe if you want them directly in your inbox! 
+That doesn't mean everything gets lost, though.
```
ravindk89: I was chewing on this - feel free to discard entirely - but I _think_ we can be more explicit about just how prohibitively expensive this is.

I also don't think we need to call out the pros vs cons (given that we're only talking about the commonly stated pro anyways). 

```suggestion
Paying for Slack is prohibitively expensive - for the size of our community (over 10K members), the lowest available paid Slack plan would cost somewhere in the range of 5-figures per month USD. 

Our desire to keep this community free-to-access means that the Write the Docs Slack will remain on the free plan. We have no plans to change or revisit this topic in the foreseeable future. For those members who would like to contribute voluntarily, we would prefer to direct those finances to other programs that benefit the wider community, such as the yearly Write the Docs conferences.

There have been numerous discussions both internally and within the community on the upsides and downsides of a finite history - and we are certain there will be more in the future. We respect and welcome additional input and ideas. It's as a result of those discussions that we have discovered that a finite history does not mean that everything gets lost.
```

ravindk89 (commented):
Feel free to toss this, it was rattling around my head while reading through this.

plaindocs (approved):
Nice work

docs/slack.rst
ddbeck: I like what you're thinking here, Ravind, but I realized that adding more text here hides the actual thing we want to achieve: for the community to understand the limitation and to spread what they've learned beyond Slack.

I moved the "Why don't we pay for Slack?" question to it's own note box. That said, I didn't follow this suggestion fully because, while it's all true, I think there's some value in being assertive and not explaining ourselves too much.

janine-c: I think what you've both come up with looks really great! I think the "5 figures USD per month" thing is useful to highlight - a lot of people who ask us do so with the assumption that it can't be _that_ expensive, and a ballpark number might be helpful to head off those repeated inquiries.

docs/slack.rst
```diff
@@ -27,16 +27,26 @@ Feel free to send a Pull Request to update `this page`_, if you want to include
...
+- `Save a message`__ privately in Slack.
+- Recap what you learned and share it on your own site or social media accounts. Don't forget to tag `#writethedocs`!
 
+.. __: https://slack.com/intl/en-gb/help/articles/360042650274-Save-messages-and-files-
+
+.. note::
+
+   Why don't we pay for Slack?
+   The Write the Docs Slack community is free-to-access and we aim to keep it that way.
+   A Slack subscription is prohibitively expensive for a community like ours.
```
ravindk89: Echoing @janine-c 's comment to maybe keep the estimated number here. Spitballing:

```suggestion
   A Slack subscription is prohibitively expensive for a community like ours. We have over 10K members, which at even the lowest possible per-user plan would cost 5-figures USD per month to maintain.
```

docs/slack.rst
```diff
@@ -27,16 +27,26 @@ Feel free to send a Pull Request to update `this page`_, if you want to include
...
+- Contribute a summary to one of the Write the Docs site's :doc:`/about/learning-resources`.
+- Turn a discussion into a talk proposal for :doc:`an upcoming Write the Docs conference </conf/index>`.
+- `Save a message`__ privately in Slack.
+- Recap what you learned and share it on your own site or social media accounts. Don't forget to tag `#writethedocs`!
 
+.. __: https://slack.com/intl/en-gb/help/articles/360042650274-Save-messages-and-files-
+
+.. note::
+
+   Why don't we pay for Slack?
```
ravindk89:
```suggestion
.. admonition:: Why don't we pay for Slack?
   :class: note

```

Unless I'm completely wrong or WtD's theme doesn't work this way, I *think* this should create a note admonition with "Why don't we pay for Slack" as part of the bolded card title.

ravindk89 (approved):
LGTM. Left two optional suggestions.

plaindocs:
image

and 

image

docs/slack.rst
plaindocs: :100:

ddbeck: I tried this first and it inserts `:` at the end of the title 😞

plaindocs: weird. I didn't even notice that. :/

I keep trying to use them in the conf pages well, :facepalm:

plaindocs: I guess cos `Note:` does too

plaindocs: Could reword to 

```
.. admonition:: Why we don't pay for Slack
```

ddbeck:
More admonishment and $$$ in the latest push.

ravindk89 (approved):
LGTM!

plaindocs:
:shipit:

PR #1653, opened 2022-01-19
Updating the book club page

djwfyi:
We have swapped the books being read in the book club and delayed the start date for early 2022.
This merge updates the book club page to reflect those changes.


plaindocs (approved):
:+1:

PR #1655, opened 2022-01-24
Replacing expired invite link with new link

janine-c:


ravindk89 (approved):
I assume this should do the trick since Slack said it never expires :)

PR #1673, opened 2022-03-03
Updated link to Red Hat style guide

stoobie:
1. Red Hat is an enterprise, not a project, so I moved the link appropriately.
2. The link you have is old. Technical writers at Red Hat no longer use the style guide at the previous link, although some other people at Red Hat still do, which is why it is still on line. The link I added is the current style guide we use.


plaindocs (approved):
Thanks!

PR #1674, opened 2022-03-03
POC for a better spreadsheet

plaindocs:
For the CFP review meeting, we copy-pasta a spreadsheet that looks a bit like this:

```
Median;[Title](link);Speakers;Tags
4;Talk title; Good people; Tags
3;ANother good one; Folks;
```

While the Median is useful, often you end up with 30 identically scored talks with no secondary sorting. I think it would be more useful to us to have not only Median, but within that Sort by Average. 

This gets all of the talk codes (from which you can build the URL) and titles, as well as all of the scores, associated by talk code.

I ran out of time and motivation to mash the two together, made a separate call for Speakers name info, calculate the average and median, and spit out a sorted spreadsheet.

@mxsasha I don't know if you fancy pushing it over the line? There is no rush, that is for sure. But I can also have another poke when I have a moment.

### DIFF
--- a/docs/_scripts/pretalx2cfpsheet.py
+++ b/docs/_scripts/pretalx2cfpsheet.py
@@ -11,13 +11,7 @@ sys.path.insert(1, '../_ext')
 
 import os
 from collections import OrderedDict
-from utils import slugify
 import requests
-from ruamel import yaml
-from ruamel.yaml.representer import RoundTripRepresenter
-import markdown
-
-SPEAKER_IMAGE_PATH = '../_static/img/speakers/'
 
 def get_review_scores(pretalx_slug):
     if not os.environ.get('PRETALX_TOKEN'):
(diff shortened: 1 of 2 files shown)


mxsasha:
Looks neat, I can pick it up :)

plaindocs:
You know what else would be awesome? 

A spreadsheet of all speakers (including folks who were not accepted), sortable by conference, year or status. 

```
Speaker, Talk, Conference, Year, Accepted status
```

kinda thing

plaindocs:
@mxsasha reckon you'll be able to look at this before the end of the month?

mxsasha (approved):

docs/_scripts/pretalx2cfpsheet.py
```diff
@@ -0,0 +1,127 @@
...
+            'wtd-portland-2020',
+            'write-the-docs-portland-2021',
+            'wtd-portland-2022',
+            'wtd-prague-2020',
+            'write-the-docs-prague-2021',
+            # 'wtd-prague-2022',
+            'write-the-docs-australia-2020',
+            'wtd-australia-and-india-2021',
+        ],
+        output_file='wtd-portland-2022-reviews.csv',
```
plaindocs: Wanna use pretalx_slug here? And add  `*-eviews.csv` to gitignore just in case?

plaindocs (commented):
LGTM

PR #1681, opened 2022-03-22
Fix #1311 - Limit file name length by limiting talk slug length

### LINKED ISSUE #1311: File name length
Two PRs I submittied (#1305  and #1306) had problems with three deleted files.
I suspect this was due to my OS (Linux Mint 18.3) not liking the length of the file names and refusing to download the files from the repo.

I suggest that we work to limit the number of characters in file names. If I recall correctly, Windows has issues with file name lengths > 255 characters. 

And we don't have control over where a contributor stores the files on their local PC, so we should aim for a file name length that accounts for that.

mxsasha:
This also trims the slugs for some Portland 2022 talks, but those should only have anchor links, if anything at all.
This will be reflected in the video archive once we produce that.

Historical long filenames for videos are unchanged because that would break external links.

Our longest historical slug is no-community-members-were-harmed-in-the-making-of-this-doc-sprint-how-we-ran-a-48-hour-event-to-collect-community-wisdom-into-a-guidebook-for-newsroom-developers-ryan-pitts-lindsay-muscato which is now 161 characters, and will be limited now to 91 (we also include one speaker name in the slug since last year).

### DIFF
--- a/docs/_data/portland-2022-sessions.yaml
+++ b/docs/_data/portland-2022-sessions.yaml
@@ -1,6 +1,6 @@
 - title: 'Measuring Documentation Success in Open Source: Findings from Google''s
     Season of Docs Program'
-  slug: measuring-documentation-success-in-open-source-findings-from-google-s-season-of-docs-program-kassandra-dhillon
+  slug: measuring-documentation-success-in-open-source-findings-from-google-s-season-of-kassandra-dhillon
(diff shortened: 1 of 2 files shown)


mxsasha:
Needs #1682 to be fixed to build

plaindocs (approved):
LGTM

PR #1683, opened 2022-03-24
Empty schedule

plaindocs:
OK folks, time to decide how we're going to do this. Current PR status is Portland 2021 (14 talks) stripped of talk slugs.

We're going down to 12 talks, consolidating Q&A to every two talks. Do we want to lop them off Tuesday end of day? 
What did we decide on Writing day? Job fair? Sponsored X/Y/Z?


plaindocs:
This build will fail until we match up the session slugs with real info.

plaindocs:
I'm opening this up to discussion early because I want to avoid doing a bunch of work that needs redoing later like we've sometimes had in previous schedules. And this one is more different.

plaindocs:
OK folks, we've written in enough protective code top protect the schedule page that I can't easily see temporary schedule changes like the ones I'm making above. :man_facepalming: 

Editing schedules like this is still super not fun. 

Take a look at **day 1** and suggest any changes inline.

docs/_data/portland-2022-schedule.yaml
```diff
@@ -0,0 +1,96 @@
...
+  - time: '12:30'
+    title: Conference Intro and Tour with the Welcome Wagon
+  - time: '5:00'
+    title: End of Writing Day
+
+talks_day1:
+  - time: '8:00 <a href="https://time.is/PT">Pacific</a>'
+    title: Platform opens
+  - time: '8:30'
+    title: Welcome Wagon intro to WTD
```
thatdocslady:
```suggestion
    title: Welcome Wagon conference introduction and platform tour
```

docs/_data/portland-2022-schedule.yaml
```diff
@@ -0,0 +1,96 @@
...
+  - time: '10:00'
+    title: '5 minute break'
+  - time: '10:05'
+    slug: TBD 2/2
+  - time: '10:35'
+    title: '5 minute break'
+  - time: '10:40'
+    title: 'Two talk Q&A'
+  - time: '10:55'
+    title: '5 minute break'
```
thatdocslady: If we extend this break to be the morning snack break it means that lunch can be pushed to 12:30 pretty safely imo... thoughts?

docs/_data/portland-2022-schedule.yaml
```diff
@@ -0,0 +1,96 @@
...
+    slug: TBD 1/2
+  - time: '11:30'
+    title: '5 minute break'
+  - time: '11:35'
+    slug: TBD 2/2
+  - time: '12:05'
+    title: 'Two talk Q&A'
+  - time: '12:20'
+    title: 'Lunch Break'
+  - time: '1:20'
```
thatdocslady: Lightning talks should be here, no?

thatdocslady (changes requested):
Looks great, started commenting and then got stuck after lightning talks because my brain broke when trying to calculate the timings =/

plaindocs:
> Looks great, started commenting and then got stuck after lightning talks because my brain broke when trying to calculate the timings =/

Yeah, it's a real headache.

plaindocs:
I poked at editing the code to make a simple schedule display, but didn't have the brain spoons for it.

docs/_data/portland-2022-schedule.yaml
```diff
@@ -0,0 +1,98 @@
...
+  - time: '11:40'
+    title: '5 minute break'
+  - time: '11:45'
+    slug: TBD 2/2
+  - time: '12:15'
+    title: 'Two talk Q&A'
+  - time: '12:30'
+    title: 'Lunch Break'
+  - time: '1:30'
+    title: Lightning Talks (30 mins)
```
thatdocslady, line 48: Rose asked if we could have a break after lightning talks, 5 min should probably be enough

docs/_data/portland-2022-schedule.yaml
```diff
@@ -0,0 +1,98 @@
...
+  - time: '2:30'
+    title: '5 minute break'
+  - time: '2:35'
+    slug: TBD 2/2
+  - time: '3:05'
+    title: '5 minute break'
+  - time: '3:10'
+    title: 'Two talk Q&A'
+  - time: '3:25'
+    title: Day 1 wraps up
```
thatdocslady, line 62: Would it be easier to combine this with the final Q&A of the day? It's not a long update and usually done live, so it might help to consolidate things for people

thatdocslady (commented):
Round 2 - nearly there!

docs/_data/portland-2022-schedule.yaml
plaindocs: Yeah, but it makes it a 14:05 start. We could just cut one talk instead?

docs/_data/portland-2022-schedule.yaml
plaindocs: I reckon we're better off saying "The END" tbh, but don't feel very strongly

docs/_data/portland-2022-schedule.yaml
thatdocslady: That doesn't feel like a problem to me, lightning talks are talks innit ;)

docs/_data/portland-2022-schedule.yaml
thatdocslady: Oh that's what you mean :) I think that it's still worth calling out "end of day announcements" as a part of the Q&A, and maybe converting this to "Day 1 ends" because the term "wrap-up" has an announcementy feel about it...

plaindocs:
New code looks good @mxsasha 

I definitely don't like the 5 min breaks in the schedule. :trollface: 

## This year

image

## Last year

image

mxsasha:
* The config now supports `flagscheduleincomplete` which makes a best effort of [rendering a schedule](https://writethedocs-www--1683.org.readthedocs.build/conf/portland/2022/schedule/). It allows slugs that are unknown, and allows talks to be missing from the schedule. It is helpful for playing, but **important that we disable this flag once the schedule is filled** because these checks protect us from mishaps in config or code.

* Schedule converted to the nice time and/or duration format that we introduced last year. I used duration everywhere, and followed the schedule as it was in this PR alrady. Some durations seem inconsistent now, lightning talks are 30 minutes spacing on day 1, 35 on day 2, and we have talk slots of 30, 35, 40 and 45 minutes. But don't know if that was on purpose. At least it's explicit.
 
* Because it might not be immediately obvious: you can have a duration of 0:00 as a sort of "marker" in the schedule, e.g. how we start the job fair after the first talk, and at the same time start the second talk - because the zero duration of the job fair means the time does not shift for that schedule entry. Also means that job fair ends is automatically aligned to start of lunch, for example.

plaindocs:
Sounds good - I only updated day 1 because I didn't want to make a bunch of changes twice. Day 2 coming soon.

plaindocs:
Looks like this https://writethedocs-www--1683.org.readthedocs.build/conf/portland/2022/schedule/#conference-talks

docs/_data/portland-2022-schedule.yaml
ericholscher: I don't think we should cut more of the lightning talks. We already made the conference shorter. We're already using lots of weird start times, I think lunch is the main one to keep on a round number if possible.

docs/_data/portland-2022-schedule.yaml
ericholscher: I feel like the end of the day is always awkward. I don't think we have a great option, and people are always confused.

docs/_data/portland-2022-schedule.yaml
```diff
@@ -0,0 +1,116 @@
...
+  - time: '17:00'
+    title: "<b>Social Event</b>"
+  - time: '19:00'
+    title: "<b>Social Event Ends</b>"
+
+talks_day2:
+  - time: '8:00'
+    title: Platform opens
+  - duration: '0:00'
+    title: "<b>Unconference Starts</b>"
```
ericholscher:
```suggestion
```

Unconf is only in the afternoon.

ericholscher:
This looks great overall. I think the breaks are good, we just need to clean up the display. 👍

plaindocs:
Give the breaks a different CSS class in the template, and reduce the visual importance or some such. But it's more or less beyond my CSS.

ericholscher:
@plaindocs Yea, I can futz with it, once we have the structure in place.

plaindocs:
@ericholscher wanna prioritise futzing with the CSS on this? I'll merge this, then open a new PR putting in the speaker details (just so it's less messy for them to look at) over the next couple of days.

mxsasha (approved):

mxsasha:
Just to confirm, once we merge this, [this schedule](https://writethedocs-www--1683.org.readthedocs.build/conf/portland/2022/schedule/) including "slug incomplete" goes live but is not announced :)

(as in, the schedule build with flagscheduleincomplete isn't secret or anything, just less strict)

plaindocs:
@mxsasha yeah, though I reckon I'll flip `flagscheduleincomplete` back to false before we merge, and then make the new PR with real data.

docs/_data/portland-2022-config.yaml
```diff
@@ -230,7 +231,8 @@ flagticketsonsale: True
 flagsoldout: False
 flagspeakersannounced: True
 flagrunofshow: False
-flaghasschedule: False
+flaghasschedule: True
+flagscheduleincomplete: True
```
plaindocs:
```suggestion
flagscheduleincomplete: False
```

docs/_data/portland-2022-config.yaml
```diff
@@ -230,7 +231,8 @@ flagticketsonsale: True
 flagsoldout: False
 flagspeakersannounced: True
 flagrunofshow: False
-flaghasschedule: False
+flaghasschedule: True
```
plaindocs:
```suggestion
flaghasschedule: False
```

Although for the build not to break, I also have to turn the schedule off?

mxsasha: Correct. If you are building a schedule, it must be complete, except if the incomplete flag is set. This may not be the most optimal config, but I wanted a solution fast before we ended up with a dozen google sheets again ;)

PR #1684, opened 2022-03-26
Speaker updates

plaindocs:
Adding a new speaker to the AWS talk.


docs/_data/portland-2022-sessions.yaml
```diff
@@ -102,7 +102,7 @@
   - name: Kate Mueller
     slug: kate-mueller
     twitter: 
-    website: Can I add this later? Site is under development...
+    website: http://www.livewithsass.com/
```
plaindocs, line 105: LOL

mxsasha (approved):

docs/_data/portland-2022-sessions.yaml
mxsasha: That's what we get for not doing external link checking ;)

plaindocs: Maybe at least check for `^http`? That way we can manually fix up

PR #1691, opened 2022-04-04
Add a couple new sponsors

ericholscher:


mxsasha (approved):
RTD build failed at first, seems to have been random, force push to rebuild fixed it

PR #1695, opened 2022-04-05
Mention pinging staff as part of newsletter

bethaitman:


ericholscher (approved):
Relevant 💯

plaindocs:
@bethaitman as `docs/organizer-guide/newsletter/template.rst` ~is only 4 lines long,~ should we just include it in a code block in the guide? Or would that be more annoying.

PR #1713, opened 2022-05-03
Add corporate concierge tickets for Prague 2022

mxsasha:


mxsasha:
https://writethedocs-www--1713.org.readthedocs.build/conf/prague/2022/tickets/

docs/conf/prague/2022/tickets.rst
```diff
@@ -100,6 +100,21 @@ Purchase this ticket if you are currently enrolled as a student, or don't curren
 
 .. class:: ticket
 
+**Corporate Concierge Tickets** *{{tickets.concierge.price}}*
+------------------------------------------------------
+
+We offer a corporate concierge service if your company is unable to follow our regular ticket sales process through the website.
+We can offer payment by invoice, process purchase orders, bank transfers, fill in supplier registration forms, and offer other support.
+Note that payment must still be received in full before the start of the conference.
```
thatdocslady: Should we add something extra-explicit like "we reserve the right to cancel registrations for unpaid tickets" or something like that? To make it clear that if they don't pay then they won't be able to log into the event.

mxsasha: I think a reminder during the process will fit better. Once we try this, if it is actually used enough, I can come up with some templates in helpmonks.

docs/_data/prague-2022-config.yaml
```diff
@@ -67,6 +67,8 @@ buttons:
     #   link_absolute: https://www.youtube.com/playlist?list=PLZAeFn6dfHpkBJAPYFrob6gqdiBuePwGJ
 
 tickets:
+  concierge:
+    price: €350
```
ericholscher: This seems a bit steep. I wonder about lowering the price, but having a limit of at least 3 tickets? I'm also fine with pricing higher but not having a limit, but seems more likely that these folks will want multiple tickets.

ericholscher (approved):
Let's see if we get any bites.

docs/conf/prague/2022/tickets.rst
ericholscher: Yea, I don't think we need stronger language here. I'm also mostly fine allowing payment after the date, but it's a good policy to start with.

docs/_data/prague-2022-config.yaml
thatdocslady: Is it technically possible to have a minimum ticket amount on Tito? If so, then we can probably do 250 with a minimum of 3-4 tickets..

mxsasha: These don't go through tito, that's the point ;)

€ 250 with a 3 ticket minimum? Don't want to make it too attractive either, because some corporates can use the normal tito process with more effort on their side, we want to keep that as the ideal option because it scales well.

ericholscher: Yea, I was just looking at the Djangocon tickets, and there's are $200 more, but that's only ~25% of the total cost. I think a minimum of 300 EUR for dealing with it (100 EUR * 3) feels like plenty, and makes it a bit less overwhelming price-wise?

ericholscher (approved):

PR #1721, opened 2022-05-15
Updates to WW guide

canncrochet:
Updates including removing Networking, adding My Agenda, View Event Schedule button, and About You convo topics.


canncrochet:
Not needed, this change included in PR 1722

PR #1724, opened 2022-05-16
Try using opengraph for nicer embeds

ericholscher:
* Before: https://www.opengraph.xyz/url/https%3A%2F%2Fwww.writethedocs.org%2Fnewsletter%2F/
* After: https://www.opengraph.xyz/url/https%3A%2F%2Fwritethedocs-www--1724.org.readthedocs.build%2Fnewsletter%2F/

Unfortunately doesn't work for conference pages, because we override the templates :(


plaindocs (approved):
Looks good to me.
Can't we re-override the templates back on the conf pages?

ericholscher:
@plaindocs yea, just not sure how it's injecting the values.

ericholscher:
Added 👍

plaindocs:
Could use a title (perhaps with a default) and a little tweaking of the spacing if you're up for it.

image

ericholscher:
Yea, that page and many others probably needs an explicit override. Probably more work than I wanted to take on with this PR heh

plaindocs:
If it is significantly better than what we have we could merge now and update later.
Better would be to fix the title, and leave the spacing for another day.

plaindocs:
Direct link to the Opengraph conference preview: https://www.opengraph.xyz/url/https%3A%2F%2Fwritethedocs-www--1724.org.readthedocs.build%2Fconf%2Fportland%2F2022%2F/

docs/conf/portland/2022/index.rst
```diff
@@ -1,5 +1,8 @@
 :template: {{year}}/index.html
 :banner: _static/conf/images/headers/{{ shortcode }}-group.png
+:og:title: Home - Write the Docs {{ name }} {{ year }}
```
plaindocs, line 3: Weird this doesn't seem to be working.

```
  <meta property="og:title" content="<no title>" />
```

Wonder if we can do something like a default site title, and then overrride it IF there is a page title. Which would give us a boring WTD on every page till we fix the in page thing, but much better than no-title

plaindocs: Which is probably what you're trying to do here? 
https://github.com/writethedocs/www/pull/1724/files#diff-edc5891e3455bb5cd05af781bd73215c81f277e7a78a6350e2198621d6a898c8R46

ericholscher:
Want to try and fix this again -- the busted previews are annoying me again :)

ericholscher:
I think this should be ready again.

ericholscher:
Last thing we need is a better logo photo, but alas.

PR #1726, opened 2022-05-17
green bar button alignment

goatlady:
changed to centered at > 640px width


plaindocs:
@mxsasha do you see any reason not to remove all of the Gulp tooling?

plaindocs:
@goatlady lemme know if you need me to run the SaSS converter again.

goatlady:
> @goatlady lemme know if you need me to run the SaSS converter again.

ah, if it's not automated then yes please

goatlady:
@plaindocs if you would like to check https://writethedocs-www--1726.org.readthedocs.build/conf/portland/2022/ on mobile, the sponsor logos are not cut off now - there is still a sideways scroll but that's something that's happening over the entire site, I haven't worked out what that is (yet 👺). And on desktop, the "sponsor" and "buy a ticket" buttons are now centered not right-aligned.

plaindocs:
I think these are two slightly different issues?

goatlady:
> I think these are two slightly different issues?

Yeah - the entire site has something global causing a small amount of sideways scrolling on mobile, it's on every page that I've checked. But the fix I put in constrains the logos to that "slightly wider" boundary (where they are still visible) whereas before they were cut off.

plaindocs (approved):
Let's merge for now and revisit?

plaindocs:
Thanks @goatlady !

PR #1728, opened 2022-05-19
Last (?) speaker updates

plaindocs:
@ericholscher @mxsasha we need to be real careful updating the `session.yaml` by hand, cos it'll be blown away next time we update from Pretalx. You'd think that wouldn't happen at this late stage, but...

The alternative would be to add the session to Pretalx manually.


mxsasha:
> @ericholscher @mxsasha we need to be real careful updating the `session.yaml` by hand, cos it'll be blown away next time we update from Pretalx. You'd think that wouldn't happen at this late stage, but...
> 
> The alternative would be to add the session to Pretalx manually.

At this point, I'd say let's leave it and do careful manual updates. Feels like we are just as likely to break something by automating it now.

mxsasha (approved):

plaindocs:
Thanks @mxsasha

PR #1729, opened 2022-05-19
Tweak a couple logos

ericholscher:


mxsasha:
I also updated Hopin with the new logos.

PR #1737, opened 2022-05-22
Make intro 25 mins

ericholscher:
The video is that long,
so we can put an explicit break in the schedule.


plaindocs (approved):

mxsasha:
Modified in hopin too.

PR #1764, opened 2022-07-14
htmlproofer upgraded major version number

plaindocs:
And changed some stuff around. 
This is a first pass to make it work. 

I'm not sure what happened to the internal domain. 

CC @mxsasha @CollierCZ

### DIFF
--- a/docs/_static_html/conf/eu/2016/cfp/index.html
+++ b/docs/_static_html/conf/eu/2016/cfp/index.html
@@ -114,7 +114,7 @@
 
     
   <div class="hero hero-conf">
-    <div class="hero-logo" data-proofer-ignore>
+    <div class="hero-logo">
       <a href="/conf/eu/2016/">
         <img src="../../../../_static/img/2016/site-logo.png" />
       </a>
--- a/docs/_static_html/conf/eu/2016/cfp/index.html
+++ b/docs/_static_html/conf/eu/2016/cfp/index.html
@@ -114,7 +114,7 @@
 
     
   <div class="hero hero-conf">
-    <div class="hero-logo">
+    <div class="hero-logo" data-proofer-ignore>
       <a href="/conf/eu/2016/">
         <img src="../../../../_static/img/2016/site-logo.png" />
       </a>
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -110,4 +110,4 @@ jobs:
 
       - name: Check links
         run: |
-          htmlproofer --ignore-files "/404/,/2013/,/2014/,/2015/,/2016/,/2017/,/search\/index.html/" --allow-hash-href  --ignore-empty-alt '/.*/' --disable-external docs/_build/html
+          htmlproofer --ignore-files "/404/,/2013/,/2014/,/2015/,/2016/,/2017/,/search\/index.html/" --allow-hash-href  --ignore-empty-alt --disable-external docs/_build/html
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -110,4 +110,4 @@ jobs:
 
       - name: Check links
         run: |
-          htmlproofer --ignore-files "/404/,/2013/,/2014/,/2015/,/2016/,/2017/,/search\/index.html/" --allow-hash-href  --ignore-missing-alt --disable-external docs/_build/html
+          htmlproofer --ignore-files "/404/,/2013/,/2014/,/2015/,/2016/,/2017/,/search\/index.html/" --allow-hash-href --enforce-https=false --ignore-missing-alt --disable-external docs/_build/html
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -110,4 +110,4 @@ jobs:
 
       - name: Check links
         run: |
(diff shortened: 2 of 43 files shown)


plaindocs:
@mxsasha looks like we've got approx 2000 errors with this config, mostly plain `http://`, do you have a strong preference for any of the following:

* turn off that error :imp: (this should be an option but might not be)
* replace all `http:` with `https:` and see how many errors remain? 
* fix them manually

plaindocs:
Here are some potentially valid errors

```
For the Links check, the following failures were found:

* At docs/_build/html/conf/australia/2018/index.html:205:

  'a' tag is missing a reference

* At docs/_build/html/conf/australia/2019/index.html:208:

  'a' tag is missing a reference

* At docs/_build/html/conf/australia/2020/index.html:246:

  'a' tag is missing a reference

* At docs/_build/html/conf/australia/2021/index.html:94:

  'a' tag is missing a reference

* At docs/_build/html/conf/australia/2022/index.html:255:

  'a' tag is missing a reference

* At docs/_build/html/conf/cincinnati/2018/index.html:205:

  'a' tag is missing a reference

* At docs/_build/html/conf/portland/2018/index.html:215:

  'a' tag is missing a reference

* At docs/_build/html/conf/portland/2018/index.html:246:

  'a' tag is missing a reference

* At docs/_build/html/conf/portland/2019/index.html:218:

  'a' tag is missing a reference

* At docs/_build/html/conf/portland/2020/index.html:250:

  'a' tag is missing a reference

* At docs/_build/html/conf/portland/2021/index.html:253:

  'a' tag is missing a reference

* At docs/_build/html/conf/portland/2022/index.html:257:

  'a' tag is missing a reference

* At docs/_build/html/conf/prague/2018/index.html:219:

  'a' tag is missing a reference

* At docs/_build/html/conf/prague/2019/index.html:225:

  'a' tag is missing a reference

* At docs/_build/html/conf/prague/2020/index.html:250:

  'a' tag is missing a reference

* At docs/_build/html/conf/prague/2021/index.html:253:

  'a' tag is missing a reference

* At docs/_build/html/conf/prague/2022/index.html:259:

  'a' tag is missing a reference

* At docs/_build/html/conf/vilnius/2019/index.html:207:

  'a' tag is missing a reference
```

Which remained after brute forcing all links to `https`.

I'm out of steam for now anyway

plaindocs:
Just found this https://github.com/gjtorikian/html-proofer/blob/93ba616eb45b7ba844197fc29824995e8fcd2279/UPGRADING.md

plaindocs:
Finding out if it is possible to leave the http https://github.com/gjtorikian/html-proofer/issues/727

plaindocs:
Oh, there is a temporary option of using `sudo gem install html-proofer -v 3.19` in the GitHub action file to stick to older version as well.

plaindocs:
OK, that was an unexpected journey. There are at least a couple of band-aids in here that we should revisit:

* the HTTP thing. See #1765 
* the a tag without a link in current confs
* alt text in current confs

But for now, this turns link checking back on. :-p

docs/_static_html/conf/eu/2016/cfp/index.html
```diff
@@ -114,7 +114,7 @@
 
     
   <div class="hero hero-conf">
-    <div class="hero-logo">
+    <div class="hero-logo" data-proofer-ignore>
       <a href="/conf/eu/2016/">
         <img src="../../../../_static/img/2016/site-logo.png" />
       </a>
```
CollierCZ: Would the error go away if the image had an `alt` attribute? It seems like it's an empty link for people who don't load images, which isn't ideal for accessibility. Though I doubt there's many people going back to look at the logo from 6 years ago.

CollierCZ (approved):
Thanks for the fix! Hope the journey wasn't too demanding.

I had one question, but it's probably not a blocker either way.

docs/_static_html/conf/eu/2016/cfp/index.html
plaindocs: Ah, I was over enthusiastic in my search replace. The ignore there is intended for the bunch of `a` tags that don't have a reference, like this one: https://github.com/writethedocs/www/pull/1764/files#diff-7a114b4493925c9b9c38c932134ded5be5f69f6f371dec742f42271428f9d96cR45

We don't check for empty or missing alt tags, because we have so many, although I'd definitely be interested in fixing that, at least in current conferences. What is best practice for alt tags for images that don't mean anything and are just part of the theme?

plaindocs: I've removed the unnecessary changes, thanks.

CollierCZ: For those, an empty alt attribute:

```html
<img src="image.png" alt="" />
```
That signals that it's just decoration and is intentionally blank

plaindocs: Right, so we have the following two options for htmlproofer:

```
  p.option 'ignore_empty_alt', '--ignore-empty-alt', ' If `true`, ignores images with empty/missing alt tags (in other words, `<img alt>` and `<img alt="">` are valid; set this to `false` to flag those)'
  p.option 'ignore_missing_alt', '--ignore-missing-alt', 'If `true`, ignores images with missing alt tags'
```

And we could run htmlproofer multiple times, for example ignoring alts on everything before 2022, and another not ignoring alts but allowing empty on 2022... 

The place to start there would be in  `docs/_templates/2022/*`

PR #1765, opened 2022-07-15
Enforce https everywhere

plaindocs:
As an optional add on to #1764 I just changed everything to https, including inside SVGs and what have you, and turned on validation. :man_shrugging:


plaindocs:
I've had a quick look, and might need to revert this one

```
g.src=('https:'==location.protocol?'https://ssl':'https://www')+'.google-analytics.com/ga.js';
```

Or maybe not? It appears in a bunch of places.

ericholscher:
This seems overzealous. We should just focus on content. At first glance I noticed it broke the SVG's on the conf sites, somehow.

plaindocs:
Probably, especially if it breaks things. :-p

But we should also look at easy changes that make linting the content easier in the long run, which I'd file this under, even if this specific example is a no-go.

mxsasha:
While I do think HTTPS links are better, I see a few issues:

* It breaks the headers in SVG files (easiest to fix, just leave them)
* For all external domains, we don't know if they actually support HTTPS. We may be breaking some of the links.
* Like you already noticed, it changes the google analytics includes, and may silently be breaking other unexpected things with how broad it is now.

I could make a case for internal links, but this isn't the big win there either - that would be increasing the HSTS duration which is currently only set to 1337 seconds and impacts external and internal links in a robust way.

mxsasha (changes requested):
I think this approach is much too wide - it breaks SVGs because the header is changed (easy fix there), we don't know which of our external links support HTTPS so may be breaking all kinds of links, and indeed should not be touching analytics javascripts.

We could limit it to internal links, but then the bigger and easier win is to increase our HSTS duration which is now only 1337 seconds. That secures both internal and external inbound links in a robust way.

Issue #1774, opened 2022-08-02
html-proofer 4.2.0 -> 4.3.1 causing issues


plaindocs:
Last successful build used 4.2.0 and recent failing builds are using 4.3.1

I'm guessing they switched something out again. :/

CC @rosewms

plaindocs:
Maybe the ignore files / urls thing isn't right?

mxsasha:
It seems to ignore all arguments, it is also checking external links (see end of log)

plaindocs:
I wonder if we need to set `=true` specifically https://github.com/gjtorikian/html-proofer/pull/738/files

mxsasha:
> I wonder if we need to set `=true` specifically https://github.com/gjtorikian/html-proofer/pull/738/files

Didn't seem to work for me.

plaindocs:
Yeah, that is the thing, we just need to do it on all of them. Locally: (I also excluded `videos` cos I don't build them locally)

```
htmlproofer --ignore-files "/404/,/2013/,/2014/,/2015/,/2016/,/2017/,/search\/index.html/" --allow-hash-href=true --enforce-https=false --ignore-missing-alt=true --ignore-urls '/videos\//' --disable-external=true docs/_build/html
Running 3 checks (Images, Links, Scripts) in ["docs/_build/html"] on *.html files...


Checking 4775 internal links
Ran on 894 files!


HTML-Proofer finished successfully.


❯ htmlproofer -v
htmlproofer 4.3.1

```

PR #1776, opened 2022-08-03
Unpin Gem, add true/false to all boolean options

plaindocs:
Fix for #1774


mxsasha (approved):
Weird, thought I tried this and it didn't work. But it works in CI clearly, so 👍🏻

plaindocs:
I think, if you miss one of the options, none of them work. At least when I tried it yesterday, I only tried adding to `--disable-external`

PR #1777, opened 2022-08-03
2022 Prague schedule

plaindocs:
This is not yet confirmed with the speakers, but I've taken into account the timezones that I know about. 

- [x] Needs a blog post
- [x] Needs confirmation from the speakers (I could mail tomorrow and we can publish Monday?)

Update:

OK, now we're using the updated template, we should probably fix the two isolated talks on Tuesday, so it's 2 and 2, not 2 1 1, right?


docs/_data/prague-2022-schedule.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+talks_day1:
+- time: '9:00'
+  duration: '0:30'
+  title: Platform opens
+- time: '9:30'
+  duration: '0:30'
+  title: Conference Intro and Tour with the Welcome Wagon
+- duration: '0:40'
+  title: Write the Docs Team - Introduction to Write the Docs
```
thatdocslady: Are we not aligning this with Portland's format? We shortened the intro and renamed it to "Community update" or something, and the next talk started at :30

thatdocslady: Also, I see that the talks are still listed in sets of 3 and no explicit Q&A. Did we decide to revert this?

plaindocs: Urgh, I copied from Prague not from Portland

thatdocslady:
@plaindocs thanks for updating the schedule format! Now we just need to shift the schedule 1hr later because Europe ;)

docs/_data/prague-2022-schedule.yaml
```diff
@@ -0,0 +1,126 @@
...
+  - duration: '0:30'
+    slug: creating-documentation-for-the-african-audience-benny-ifeanyi-iheagwara
+  - duration: '0:15'
+    title: 'Closing address'
+  - duration: '0:15'
+    title: '✋ Q&A for the two previous talks'
+  - duration: '0:20'
+    title: '<b>Main stage talks done</b>'
+  - duration: '1:30'
+    slug: adhd-and-its-impact-on-the-creative-process-for-writers-tyler-smith
```
thatdocslady: Why is this after the main stage talks done? Looks like a copy-paste error

plaindocs: Yeah, that's the bit we need to reorganize, it was the Google Workshop 

> OK, now we're using the updated template, we should probably fix the two isolated talks on Tuesday, so it's 2 and 2, not 2 1 1, right?

4 talks in the morning, 2 in the afternoon with a later lunch?

plaindocs: We're probably also not running this? 

> Meetup information session - Get involved locally!

thatdocslady: Yeah let's clean up the workshop and whatever else isn't the talks that we selected, and if anything gets added back in then we'll add it :)

thatdocslady (approved):
Looks great! Now we just need the speakers to confirm and we can publish as soon as that's done :)

plaindocs:
Slightly behind with this, told speakers that we're publish **Wednesday**/

mxsasha (commented):
Should we give all items that are not a scheduled talk an emoji? Lunch break and lightning talks do have less visibility now to my eye when I look at the rendered schedule. Otherwise, 👍🏻

plaindocs:
@mxsasha I've updated those two. If you have the time you could look at Writing day, even talks, but I don't think either are blockers.

image

PR #1780, opened 2022-08-10
Fix some issues in Prague Sunday/Monday schedule

mxsasha:
Note that the move in Monday introduction shifts the entire schedule by 1 hour even though those YAML lines didn't change. Maybe we should also move up the social event? My change leaves only 30 minutes between wrap up and start of social.

[Current live](https://www.writethedocs.org/conf/prague/2022/schedule/) / [This PR](https://writethedocs-www--1780.org.readthedocs.build/conf/prague/2022/schedule/)

### DIFF
--- a/docs/_data/prague-2022-schedule.yaml
+++ b/docs/_data/prague-2022-schedule.yaml
@@ -1,9 +1,9 @@
 writing_day:
   - time: '9:00'
     title: 'Platform opens'
-  - time: '9:00'
+  - time: '10:00'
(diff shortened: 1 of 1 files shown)


ericholscher (approved):

ericholscher:
> Note that the move in Monday introduction shifts the entire schedule by 1 hour even though those YAML lines didn't change. Maybe we should also move up the social event? My change leaves only 30 minutes between wrap up and start of social.

I think an earlier social is good, but having a bit more gap might make sense. I don't feel too strongly either way.

PR #1781, opened 2022-08-15
Final Prague schedule updates 🤞

mxsasha:
Consistently has 5 minute break between two talks (or LT slot), but no break between last talk and Q&A for smoother flow.

https://writethedocs-www--1781.org.readthedocs.build/conf/prague/2022/schedule/


plaindocs (approved):
:+1:

PR #1782, opened 2022-08-17
Speaker switch for timezone reasons

plaindocs:
Confirmed by both speakers.


mxsasha (approved):
Feel free override the merge even with links-internal failing, because that will be fixed in #1781 - I'll merge that one after this one is done

PR #1784, opened 2022-08-18
Spelling fix and add ZAR to wordlist

mxsasha:


plaindocs (approved):

codespell/ignore.txt
```diff
@@ -10,3 +10,4 @@ tread
 wan
 wee
 wit
+ZAR
```
plaindocs:
```suggestion
zar
```

maybe?

mxsasha: 💯 good catch!

PR #1785, opened 2022-08-19
Update old brand name Transferwise to Wise

EDDYMENS:
Changed the name of the old brand Transferwise to Wise.


plaindocs (approved):
Thanks!

PR #1787, opened 2022-08-22
Update docs-as-code.rst

diana-lakatos:
I propose to include a link to my talk at Write The Docs Prague 2020 because I think it might be useful for community members trying to learn about docs as code. I explain the planning and implementation of our docs as code processes in detail, and share some insights about why we decided to go with this approach and how it helped us. 
Please consider if my talk is a useful addition to this page. Thank you very much!


plaindocs (approved):
Seems good to me!

PR #1793, opened 2022-08-24
Fix link to my site

CollierCZ:
Sorry I didn't notice this earlier, but the current link is to a different Aaron Collier :slightly_smiling_face:


mxsasha (approved):

mxsasha:
Thanks! It had not gone out yet, fortunately the PR was merged just after our mailing time. So the when we mail it out tonight it should be correct :)

ericholscher:
Woops, thanks for the update 👍

PR #1814, opened 2022-09-27
Simplify the landing page

plaindocs:
- trim the side bar
- trim the long link lists
- trim the text 

----
:books: Documentation preview :books:: https://writethedocs-www--1814.org.readthedocs.build/en/1814/


plaindocs:
Update

image

ericholscher:
I think this still feels a bit stark. I think pulling some content from the secondary index pages that explain what each of the major sections are is worth doing. Having just a big list of links isn't great, and I think there's definitely room for more headings and content.

plaindocs:
Yeah. i do think the best way to fix it would be with design not content, but the way forward is to get this into a mergable state, and think about iterating after.

I'll rexamine this coming week.

plaindocs:
image

docs/index.rst
```diff
@@ -8,51 +8,47 @@
...
-.. note:: You can find information and updates about the impact of COVID-19 on Write the Docs conferences and meetup events at the following links: `Conference updates <https://www.writethedocs.org/blog/write-the-docs-2016-year-in-review/>`_, `Meetup events <https://www.writethedocs.org/organizer-guide/meetups/covid-19-meetups/>`_.
+Find or advertise a job
+-----------------------
 
-Join the community
-------------------
+Work with other documentarians.
 
-Get more information on how to meet the community, learn new things, get involved, and stay in touch. We have a few sets of resources for you to start with:
+* `Find a docs related job <https://jobs.writethedocs.org/>`__
```
ericholscher:
```suggestion
* `Find a docs related job with our Job Board <https://jobs.writethedocs.org/>`__
```

I think a bit more context on where some of these links go is important, otherwise it's a kind of confusing call to action

ericholscher (approved):
This is a solid improvement. I still feel like we're hiding a lot on this page, but that's probably fine? 

There's more room for improvement here, but this is good to ship I think 👍

docs/_templates/navigation.html
```diff
@@ -10,23 +10,6 @@ <h3>Job Postings</h3>
         </ul>
 </div>
 
-<div id="toc">
-
-    <h3>Site Content</h3>
-
-    {{ toctree(collapse=False, includehidden=theme_sidebar_includehidden) }}
```
ericholscher, line 17: I do think this will make it so some of our content isn't linked anywhere and can't be indexed by Google. I think we probably want to add a page that is a full TOC that is linked from somewhere?

plaindocs: Good point, I've added that in e38f6920 and linked it from the footer

ericholscher:
💯 @plaindocs 🎉

PR #1822, opened 2022-10-03
Fix #1819 - Run sphinx in quiet mode in CI

### LINKED ISSUE #1819: Improve CI output readability
Sometimes the error causing a CI failure is right at the end, but in some cases it is earlier and we produce too much logs to make it findable. There are limits, but we can certainly make some improvement with little effort. See https://github.com/writethedocs/www/actions/runs/3150239173/jobs/5122797428 as an example.

mxsasha:
----
:books: Documentation preview :books:: https://writethedocs-www--1822.org.readthedocs.build/en/1822/

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,7 +78,7 @@ jobs:
       uses: rickstaa/sphinx-action@master
       with:
         docs-folder: "./docs"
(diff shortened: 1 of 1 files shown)


ericholscher (approved):

plaindocs:
:tada:

PR #1816, opened 2022-09-27
Add Portland 2023 data

ericholscher:
This is just a copy of the 2022 Prague site.

----
:books: Documentation preview :books:: https://writethedocs-www--1816.org.readthedocs.build/en/1816/

PR #1815, opened 2022-09-27
Announcement for 2023

mxsasha:
**Draft - dates not confirmed with venues**

Todo:

- [ ] Add the Atlantic mailchimp segment (or some way for mailing list)

https://writethedocs-www--1815.org.readthedocs.build/blog/2023-conference-announcements/

PR #1823, opened 2022-10-03
Add Atlantic 2023 landing page

mxsasha:
- [ ] Update colors to be correct?
- [ ] ~~Use a new header image on the pages - something with sea? Means no community view, but may be better than reusing Prague photos everywhere.~~
- [ ] ~~Update all the times we mention in the config YAML? Not sure if they are displayed in landing already.~~
- [ ] Make sure the atlantic email works
- [ ] Make sure we have an atlantic mailing list

This currently conflicts in `_templates/2023` with #1816 - once we merge one, we need to fix the other PR. There is no news post, as that's going to be in #1815.

I also moved the 2021 configs to the past.

----
:books: Documentation preview :books:: https://writethedocs-www--1823.org.readthedocs.build/

### DIFF
--- a/docs/_data/atlantic-2023-config.yaml
+++ b/docs/_data/atlantic-2023-config.yaml
@@ -1,0 +1,194 @@
+name: Atlantic
+
+shortcode: atlantic
+year: 2023
+city: Atlantic
+local_area: Atlantic
+area: Atlantic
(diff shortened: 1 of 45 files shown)


mxsasha:
Folded into #1835

PR #1828, opened 2022-10-19
Australia 2022: add schedule and sponsor

swapnilogale:
----
:books: Documentation preview :books:: https://writethedocs-www--1828.org.readthedocs.build/en/1828/


plaindocs:
Yeah, so in 2022 for the other confs we include a shared schedule file like 

https://github.com/writethedocs/www/blob/main/docs/conf/portland/2022/schedule.rst

which includes 

https://github.com/writethedocs/www/blob/main/docs/include/conf/schedule.rst

If we do that for Australia 2022 it breaks with an `(exception: 'dict object' has no attribute 'day_three')` error to the different formats of the conference.

Probably the best fix here is to copy `include/conf/schedule.rst` to `include/conf/schedule_au.rst`, replace `conf/australia/2022/schedule.rst` to include that file, and delete the contents of the new  `include/conf/schedule_au.rst` till it works. 

@swapnilogale lemme know if that makes sense, or if you need a hand with any of it?

@mxsasha anything to add?

mxsasha:
> @mxsasha anything to add?

Sounds reasonable. I also do not immediately see an easy way to fix the template to work for both two and three day conferences.

plaindocs (approved):
This looks ok to me. I haven't exhaustively been through the schedule or links though.

PR #1834, opened 2022-10-28
Remove phone numbers from the COC

ericholscher:
These have never been used, and I'm getting weird texts that know who I am, and trying to reduce surface area.

If we want a number folks can text/call, we should probably setup Twilio or something single-use for the conf, or probably an email -> text system.

----
:books: Documentation preview :books:: https://writethedocs-www--1834.org.readthedocs.build/en/1834/

### DIFF
--- a/docs/include/conf/coc/au-coc.rst
+++ b/docs/include/conf/coc/au-coc.rst
@@ -1,11 +1,9 @@
 **Swapnil Ogale**
 
 * swapnilogale@gmail.com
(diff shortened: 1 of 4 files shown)


mxsasha:
> If we want a number folks can text/call, we should probably setup Twilio or something single-use for the conf, or probably an email -> text system.

Kind of what I do, it's a dedicated number that just mails me a voicemail out of conference times, forwards to mobile during confs. But removing seems practical.

PR #1845, opened 2022-11-14
Remove mailchimp validation JS that submits form via AJAX

ericholscher:
This was breaking for users,
and seems unnecessary.
We can just submit the form.

----
:books: Documentation preview :books:: https://writethedocs-www--1845.org.readthedocs.build/en/1845/


plaindocs (approved):
Makes sense

PR #1859, opened 2022-12-02
Remove ATL from Quorum East

rosewms:
Group no longer has an organizer, removing from East Coast Quorum

----
:books: Documentation preview :books:: https://writethedocs-www--1859.org.readthedocs.build/en/1859/


plaindocs (approved):
LGTM

PR #1878, opened 2023-01-13
Remove NC TRP (Durham) meetup

rosewms:
Meetup is no longer active or a part of Quorum. Needs to be removed from the quorum section in a follow up PR.

----
:books: Documentation preview :books:: https://writethedocs-www--1878.org.readthedocs.build/en/1878/


rosewms:
Confirmed that info was already removed from the Quorum section.

barbaricyawps (approved):
Thanks for doing this since NC seems to have closed due to lack of interest.

PR #1892, opened 2023-02-07
Fix #1890 - Rename quorum page to read as markdown

### LINKED ISSUE #1890: Fix up broken formatting on Quorum page
The Quorum page is real broke https://www.writethedocs.org/organizer-guide/meetups/quorum/ , but it might well just be case of marking it as markdown or something. 

image

Tag me in if you need me, but I'm a bit swamped with CFP at the mo.

mxsasha:
----
:books: Documentation preview :books:: https://writethedocs-www--1892.org.readthedocs.build/


plaindocs (approved):
Thanks @mxsasha looks good!

CC @barbaricyawps

PR #1936, opened 2023-04-26
Fix script to return questions

plaindocs:
Took me longer than it should have to notice. 

----
:books: Documentation preview :books:: https://writethedocs-www--1936.org.readthedocs.build/


mxsasha (approved):
I do notice this still says "pronounce" instead of pronouns? Don't know if/when we fixed that in pretalx.

plaindocs:
That's not an error, that's the question for how you pronounce your name. :-p

PR #1966, opened 2023-05-23
Fix missing line breaks in lists on the "Opportunity grants" pages

hkosova:
https://www.writethedocs.org/conf/atlantic/2023/opportunity-grants/#are-you-part-of-a-marginalized-or-underrepresented-group-in-tech

I guess these `*` in the middle of list items are supposed to be separate items:

I changed all copies of the "Opportunity grants" page (including past conferences, just in case) so that future copies won't have this typo.

----
:books: Documentation preview :books:: https://writethedocs-www--1966.org.readthedocs.build/


plaindocs (approved):
Good eyes, thank you! And thank you for taking the trouble to dig out past years as well. :heart:

PR #1972, opened 2023-06-12
Fix atlantic news post

mxsasha:
As `City` does not exist, this now says "speakers for our first virtual conference"

----
:books: Documentation preview :books:: https://writethedocs-www--1972.org.readthedocs.build/


plaindocs:
Gah. Can we configure jinja to warn on empty template variables? 
Without doing so everywhere in the template?

plaindocs (approved):

PR #1974, opened 2023-06-13
Fix #1973 - Change speaker simple list to use speaker+talk slug

### LINKED ISSUE #1973: Bug in speaker list code?
In https://www.writethedocs.org/conf/atlantic/2023/news/announcing-speakers/#full-speaker-line-up Chris Noonan is linking to his Prague 2020 talk.

~It looks like they might have the same title, but I'm still not sure how this happened.~

@mxsasha could you take a look?

mxsasha:
This may have broken in the past too. Can't fix the mailing we sent out unfortunately.

### DIFF
--- a/docs/_templates/2021/speakers.rst
+++ b/docs/_templates/2021/speakers.rst
@@ -1,7 +1,7 @@
 {% for talk in data %}
 
 {% for speaker in talk.speakers %}
-.. _speaker-{{ speaker.slug }}:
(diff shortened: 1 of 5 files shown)


plaindocs (commented):
Speedy! Thank you.

We could add a quick link saying "looking for the 2023 talk? Click here" but it probs not worth it?

plaindocs (approved):

plaindocs:
ah, looks like there is a touch more clean up

PR #1986, opened 2023-07-11
Improve accessibility

CollierCZ:
I noticed some issues with website accessibility through some automated checks, so this should help with some of those. It doesn't address the following issues (as they seemed to require some design thinking):

* Link text not having high enough contrast with the background in conference pages
* Link text not having high enough contrast with surrounding text in any pages
* Text in the footer of conference pages (the `© 2023 Write the Docs` and two instances of `|`) not having high enough contrast with the background

Nor this as it comes from a theme:

* The alabaster theme setting a maximum viewport scale (`maximum-scale=0.9`)

Hopefully the reasons for the other changes are clear from the commit messages.

* Adding English as the language to pages
* Marking decorative images as decorative (`alt=""`)
* Adding alt text to images used as buttons
* Upgrading the UI Kit from beta to give a label to a button that didn't have it (not sure it's still idea as `aria-controls` doesn't seem to be present, but it's something)

----
:books: Documentation preview :books:: https://writethedocs-www--1986.org.readthedocs.build/

### DIFF
--- a/docs/_static/conf-legacy/2017/icomoon/demo.html
+++ b/docs/_static/conf-legacy/2017/icomoon/demo.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html>
+<html lang="en">
 <head>
     <meta charset="utf-8">
     <title>IcoMoon Demo</title>
--- a/docs/_templates/2018/base.html
+++ b/docs/_templates/2018/base.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
     <head>
         <meta charset="utf-8">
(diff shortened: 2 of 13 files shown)


plaindocs (approved):
Nice one @CollierCZ 

You might be able to turn checks for alt text on in one of the linters after this? Or I can poke around and see where to do it?

CollierCZ:
I don't think I know enough about the linting here to be effective at that. If I have time, I can look into it, but definitely no promises.

plaindocs:
Fair enough, don't worry about it then. Either I'll get to it or I won't. :-)

PR #1982, opened 2023-07-07
WTD Atlantic 2023 possible schedule

plaindocs:
- [x] What starting time
- [x] Split out Q&A? 
- [x] No social
- [x] We should add a few more columns of timezones 
- [x] meal breaks? what to do?
- [x] Make sure blog post date is correct
- [x] Update shirt cutoff date (@thatdocslady)

----
Preview 🕐:  https://writethedocs-www--1982.org.readthedocs.build/conf/atlantic/2023/schedule/
https://writethedocs-www--1982.org.readthedocs.build/conf/atlantic/2023/news/announcing-schedule/

PR #1991, opened 2023-07-24
Writing Day blog post - OSS call to action

rosewms:
----
:books: Documentation preview :books:: https://writethedocs-www--1991.org.readthedocs.build/

### DIFF
--- a/docs/conf/atlantic/2023/news/writing-day-cta.rst
+++ b/docs/conf/atlantic/2023/news/writing-day-cta.rst
@@ -1,6 +1,6 @@
 :template: {{year}}/generic.html
 
-.. post:: Jul 26, 2023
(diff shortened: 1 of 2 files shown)


mxsasha:
This is great, but we're just preparing the schedule post in #1982, so I think we should publish this a week or two after the schedule?

mxsasha:
I have included the google form link from this PR already in the [draft schedule blog post](https://writethedocs-www--1982.org.readthedocs.build/conf/atlantic/2023/news/announcing-schedule/), as it mentions writing day too.

rosewms:
@mxsasha if we could do the week after the schedule that would be ideal!

mxsasha (approved):

PR #1995, opened 2023-07-27
Add post-channel creation steps to Slack page

ddbeck:
Previously undocumented, this PR explains what happens _after_ we create a channel in Slack.

----
:books: Documentation preview :books:: https://writethedocs-www--1995.org.readthedocs.build/


plaindocs (approved):

plaindocs:
Might be worth supplying an actual example of an announcement, either real or artificial?

janine-c (approved):
:shipit:

ddbeck:
> Might be worth supplying an actual example of an announcement, either real or artificial?

Good idea. I added a joke one, though we probably ought to replace it with something real (after the next time it happens, I guess).

docs/slack.rst
```diff
@@ -168,6 +168,12 @@ If you still can't find a suitable channel, you can propose a new channel. To pr
 3. Post in **#meta** requesting a channel. Be sure to share your proposed channel name, description, and to tag your cosponsors.
 
 The Write the Docs moderation team will review your proposal.
+If your proposal is approved, then you and your consponsors must:
+
+#. Seed discussion in the new channel.
```
ravindk89, line 180:
```suggestion
#. Make a best-effort to seed and maintain a cadence of conversation.
    For example, you can post questions once a week related to the channel topic and encourage others to participate, similar to how **#learn-tech-writing** contributors have done book clubs.
```

ravindk89: I'd like to set the bar to be "If you're going to start this up, you need to do more than show up at the beginning"

docs/slack.rst
```diff
@@ -168,6 +168,12 @@ If you still can't find a suitable channel, you can propose a new channel. To pr
...
+   For example, ask an open-ended question, share a link to discuss, or post a hot take.
+#. Make an announcement in **#general** and channels related to the new channel's topic.
+   For example, if there's a new a channel for underwater basket weaving, then make an announcement like "Join us in #underwater-basket-weaving to discuss techniques, materials, and what to store in all these baskets."
```
ravindk89, line 184: Do we add down here something to the effect of "Channels which remain largely empty or quiet may be closed by the Moderation team."

ravindk89:
All nits, nothing to block an existing LGTM

docs/slack.rst
ddbeck: We've had some questions in the past from folks who were worried they were taking on a permanent commitment by suggesting a channel (e.g., "do I have to moderate the channel, if I propose it?"). I think putting in a requirement like this might scare people off more than it would spur conversation.

docs/slack.rst
ddbeck: I don't think we should make a threat we're probably never going to act on. I've never got myself together enough to do some relatively easy channel renaming. I just don't see the mod team actively _looking_ for empty channels to kill.

ddbeck (commented):
Janine had some ideas for me in chat, so I'm going to put this into draft mode and do some rework.

ddbeck:
OK, I've redone this a little emphasize _why_ we're doing this: to help conversations happen (and to head off the idea that it's a gatekeeping exercise). I've also asked channel proposers to think of their seed conversation earlier in the process, so it's not a surprise and they're likelier to remember to do it, even if we forget to ask.

docs/slack.rst
```diff
@@ -161,13 +161,26 @@ Creating channels
...
+
+#. Find two other people to cosponsor your channel idea.
+   Ask in related channels to find people who share your interest.
+#. Together, come up with a channel name and write a channel description.
+#. Post in **#meta** requesting a channel.
+   Be sure to share your proposed channel name, description, and to tag your cosponsors.
+#. Start thinking about how to seed the new channel with discussion, with an open-ended question, a link to discuss, or a hot take.
+
+The Write the Docs moderation team will review your proposal to make sure the new channel will be active, durable, and won't duplicate or create confusion with another channel.
+Whenever possible, the moderators work with channel sponsors on improvements to the proposal.
```
ddbeck, line 176: @janine-c what do you think of the revisions here? Following up on https://github.com/writethedocs/www/commit/be680b0c5f8eac050b9ab8b53c4259c19674abec#r122875538

janine-c: Love it! I wonder if "durable" will confuse people - maybe "sustainable" might make more sense? Are we thinking about making sure that the channel will stay active/useful over time here?

ddbeck: Yes, I could see that might be confusing. I was trying to capture the idea of "not temporary" (e.g., I don't want people suggesting channels that are for one-off events). A couple of alternatives might be:

- Long-lived
- Permanent
- Enduring

Or I could just omit it and go with "…active and won't duplicate…".

Any thoughts?

janine-c: Yeah, maybe we could say something like "...the new channel will remain active and won't duplicate..."? I think that even when channels feel like they'll probably meet that criteria, we can't know if they'll be permanent, but I think that phrasing hits on the idea that we don't want to have a channel that's, say, only related to a single year's hackathon or anything like that.

docs/slack.rst
```diff
@@ -161,13 +161,26 @@ Creating channels
...
+To propose a channel:
+
+#. Find two other people to cosponsor your channel idea.
+   Ask in related channels to find people who share your interest.
+#. Together, come up with a channel name and write a channel description.
+#. Post in **#meta** requesting a channel.
+   Be sure to share your proposed channel name, description, and to tag your cosponsors.
+#. Start thinking about how to seed the new channel with discussion, with an open-ended question, a link to discuss, or a hot take.
+
+The Write the Docs moderation team will review your proposal to make sure the new channel will be active, durable, and won't duplicate or create confusion with another channel.
```
ddbeck: Suggestion from https://github.com/writethedocs/www/pull/1995/files#r1306712473

```suggestion
The Write the Docs moderation team will review your proposal to make sure the new channel will remain active and won't duplicate or create confusion with another channel.
```

docs/slack.rst
ddbeck: "Remain active" I like it, done! ✅

PR #1999, opened 2023-08-03
Fix “Fork me on GitHub” ribbon

mattstein:
The global “fork me” ribbon is a broken image because the external reference returns a 403 response:

Screen Shot 2023-08-03 at 08 43 17 AM@2x

I found an SVG with the same filename, cleaned it up (with Figma + ImageOptim), and swapped it in to restore the ribbon and hopefully make it more resilient:

Screen Shot 2023-08-03 at 08 43 42 AM@2x

Tested locally to confirm It Should Work™.

----
:books: Documentation preview :books:: https://writethedocs-www--1999.org.readthedocs.build/


plaindocs (approved):
I'd not even noticed that one! 

Thanks for the fix. :-)

PR #2001, opened 2023-08-07
Slack guidelines: revise self-promotion CTA language

ddbeck:
Not everyone knows what an infomercial looks like. Here, we try to be more specific about what we're talking about.

----
:books: Documentation preview :books:: https://writethedocs-www--2001.org.readthedocs.build/


ravindk89 (approved):

plaindocs (commented):
I feel like first line and last line say pretty much the same thing, and would delete or consolidate, but don't feel strongly on the matter.

plaindocs (approved):

janine-c (approved):
Sorry about the delay on this one! Looks great - no one can accuse us of not giving enough detail this way 😄

PR #2022, opened 2023-09-08
Close ticket sales for Atlantic

mxsasha:
----
:books: Documentation preview :books:: https://writethedocs-www--2022.org.readthedocs.build/


plaindocs (approved):
Might need to change the CTA links too unless that is automatic?

mxsasha:
> Might need to change the CTA links too unless that is automatic?

Not automatic, but we already didn't have the ticket purchase buttons there.

PR #2038, opened 2023-10-07
Fix bulleted list in session description

sarahmaddox:
----
:books: Documentation preview :books:: https://writethedocs-www--2038.org.readthedocs.build/


plaindocs (approved):
Thanks @sarahmaddox !

@swapnilogale FYI I updated the source in Pretalx also so that this won't reoccur when we regenerate.

PR #2043, opened 2023-10-17
[Australia] - added STE workshop page

swapnilogale:
----
:books: Documentation preview :books:: https://writethedocs-www--2043.org.readthedocs.build/


plaindocs:
Something like this should work in `www/_/docs/_templates/2023/menu-desktop.html` and `menu-mobile.html`

```jinja
{% if shortcode == "au" %}
<li><a href="{{ pathto('conf/'+shortcode+'/'+year_str+'/workshop-ste') }}">Workshop</a></li>
{% endif %}
```

swapnilogale:
@plaindocs ready for review. I couldn't figure out how to add Part 1 and Part 2 to the workshop on the schedule page (that's because it comes from the slugs), but that's not a biggie.

plaindocs:
yeah, we'd have to mess with it. You could leave part 1 as is, and do "... part 2" as text only with no link if you wanted? 

I'd also add Workhop to the title in the schedule

swapnilogale:
> yeah, we'd have to mess with it. You could leave part 1 as is, and do "... part 2" as text only with no link if you wanted?
> 
> I'd also add Workhop to the title in the schedule

Fixed

plaindocs (approved):

PR #2052, opened 2023-11-14
Standardize on opengraph and group images

ericholscher:
There's probably a smarter way to do this with code, but I just hacked it for now.


plaindocs (approved):
Looks good.

CC @goatlady for salary survey.

PR #2055, opened 2023-11-20
Embed Slido on panel page for Au conference

flicstar:
I copied the code from the embedded Slido on the Portland Q&A page - replacing the event ID with the Australian Slido event. 

----
:books: Documentation preview :books:: https://writethedocs-www--2055.org.readthedocs.build/


plaindocs (approved):
Looks good to me

PR #2080, opened 2024-01-22
2022 survey mobile css specificity fix

goatlady:
----
📚 Documentation preview 📚: https://writethedocs-www--2080.org.readthedocs.build/


goatlady:
@plaindocs the CSS for two columns on mobile was in there, but it looks like the section name changed and I didn't catch it. All pretty now :)

plaindocs (approved):

PR #2118, opened 2024-03-28
Meetups: Add Tunis, edit Michigan, edit Inactive list

rosewms:
Several edits, plus yaml removal, new Tunis group information, updates to Archived and Inactive lists.

----
📚 Documentation preview 📚: https://writethedocs-www--2118.org.readthedocs.build/

PR #2147, opened 2024-05-08
Mtp: Add Armenia WTD group info

rosewms:
Adding the Armenia WTD group info to the Meetups page.

Note, might update the page title to "Meetups and communities" to better reflect the new platforms available.

[Merge 2118](https://github.com/writethedocs/www/pull/2118) before this PR.

----
📚 Documentation preview 📚: https://writethedocs-www--2147.org.readthedocs.build/


plaindocs:
This won't work as is, Sphinx is looking for a non existant `meetup` key, and we'll have to add some logic to handle that.

rosewms:
> This won't work as is, Sphinx is looking for a non existant `meetup` key, and we'll have to add some logic to handle that.

Any suggestions? I'd love a secondary solution because some of the groups are also creating location specific Write the Docs LinkedIn pages that I think it would be good to include.

plaindocs:
Yeah, I'll poke at the code, won't happen overnight but shouldn't be super tricky

mxsasha (approved):

plaindocs:
Hey @rosewms this is good to go! Merge at will, and lemme know if oyu have any questions?

plaindocs:
This is related to #2242 and attempts to remove the need for an actual meetup.com URL as well

docs/_ext/meetups.py
```diff
@@ -43,8 +43,9 @@ def load_meetups_by_region():
             'organizers',
             'website',
             'twitter',
+            'meetup_alt',
         ]])
-        if 'meetup' not in meetup:
+        if 'meetup' not in meetup and 'meetup_alt' not in meetup:
             raise ExtensionError('Meetup missing `meetup` key: file={0}'.format(
```
mxsasha:
```suggestion
            raise ExtensionError('Meetup missing `meetup`/`meetup_alt` key: file={0}'.format(
```

docs/_data/schema-meetups.yaml
```diff
@@ -0,0 +1,15 @@
+region: str(required=True)
+country: str(required=True)
+state: str(required=False)
+city: str(required=False)
+website: str(required=False)
+twitter: str(required=False)
+meetup: str(required=False) # I can't figure out how to validate must contain either meetup or meetup_alt
```
mxsasha: Don't think there is a way to do this in the schema

mxsasha (approved):
Good idea to add a schema for validation!

docs/_data/schema-meetups.yaml
plaindocs: Yeah, it feels like there *should* be. 

We could refactor to something like:

```yaml
source:
   # one of
   meetupcom: example
   linkedin: example
   other: example
```

and then I believe we could something along these lines https://github.com/23andMe/Yamale?tab=readme-ov-file#lists

plaindocs: OK, I've gone a ahead and done that, and it seems to work, thought it still breaks my brain a little.

plaindocs (approved):
This is mostly my PR now! But I'm happy with it :-p

docs/_data/schema-meetups.yaml
```diff
@@ -0,0 +1,21 @@
...
+source: any(include('source_meetup'), include('source_other'), required=True)
+
+---
+
+organizer-detail:
+  name: str(required=True)
+  link: str(required=False, none=False)
+
+source_meetup:
+  meetup: str(required=True, none=False)
```
mxsasha, line 18: https://www.3djake.nl/bambu-lab/a1

mxsasha, line 18: Maybe we could call this and the other one "url"? So that you only flip whether url is under source_meetup or source_other. Feels cleaner.

plaindocs: yeah, I actually went the other way to 

```
source:
  meetup:
```

and 

```
source:
   other:
```

but I'm falling into the trap of half assing things:

https://github.com/writethedocs/www/pull/2242#pullrequestreview-2373307871

rosewms:
Closing PR in light of the updated work in #2607

plaindocs:
@rosewms I think we might need to still add the Armenia file to that other PR

PR #2180, opened 2024-07-30
Resize all speaker images to be 200px wide

ericholscher:
These files were taking up a bunch of space on my disk, so figured it would be nice to resize them all down, since they're only shown as a thumbnail. 

----
📚 Documentation preview 📚: https://writethedocs-www--2180.org.readthedocs.build/


plaindocs:
Heh, "  reviewdog: post failed for vale: fail to parse diff: GET https://api.github.com/repos/writethedocs/www/pulls/2180: 406 Sorry, the diff exceeded the maximum number of files (300)."

CollierCZ:
It's an issue with the kind of check that's used. It would work with `github-pr-review`, but I think that requires API keys and maybe other differences. ((reference)[https://github.com/reviewdog/reviewdog/issues/1696])

plaindocs:
I think we can just merge this. Vale doesn't check images anyway.

CollierCZ (approved):

PR #2184, opened 2024-08-05
Clarify the purpose of starting.rst

stumblefiend:
We have several confusing pathways into multiple "getting started" types of content. Looking at the main software documentation landing page or knowing where to click from a search becomes confusing. I tried cleaning up this content and giving it a purpose, but there really looks to be only two valid scenarios the doc presents (the enterprise users section doesn't count because it links to the "interesting approaches" content instead of an actionable guide related to the topic of enterprise users.

Overall, this page may not be tangible in any meaningful way, so just wanted to hear other's thoughts...should we just delete or archive this content

### DIFF
--- a/docs/guide/starting.rst
+++ b/docs/guide/starting.rst
@@ -1,42 +1,31 @@
-============
-Starting off
-============
+=======================================================================
+Starting off: documentation best practices for developers
+=======================================================================
 
-So, you want to write the docs for your project?
+Want to write the docs for your project?
+You've come to the write [sic] place.
 
-    *Where on earth do I start?*
-
-Well, you've come to the write [sic] place.
-
-The goal here is to explain how to write docs,
-how to get the time to write docs,
-and what you should be striving for with those docs.
-A lot of projects are different, and a lot are the same.
-We will be covering the tactics for a few different groups.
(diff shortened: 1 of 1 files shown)


docs/guide/starting.rst
```diff
@@ -1,53 +1,28 @@
-============
-Starting off
-============
+=======================================================================
+Starting off: documentation best practices for developers
+=======================================================================
 
-So, you want to write the docs for your project?
+Want to write the docs for your project?
+You've come to the write [sic] place.
```
ericholscher: I think [sic] is mostly for text copied from elsewhere? Not 100% sure tho. 

```suggestion
You've come to the *write* place.
```

ericholscher (approved):
I think this is a useful "getting started" content, but could also be a top-level page in the guide for this content, instead of a subpage? 🤷

docs/guide/starting.rst
plaindocs: Right, a kinda "we know this is wrong, but didn't fix it cos it's a direct quote.

stumblefiend:
I think this PR is good enough for now and should be an improvement over what was there originally.

PR #2188, opened 2024-08-08
Add speaker pic

plaindocs:
Couple of thoughts cropped up here:

Should we automate image optimization with an action?

----
📚 Documentation preview 📚: https://writethedocs-www--2188.org.readthedocs.build/


mxsasha:
> Should we automate image optimization with an action?

Sure, why not. At least for new images, we don't want to rewrite all images all the time.

mxsasha (approved):

plaindocs:
Maybe something like this? https://github.com/calibreapp/image-actions

PR #2229, opened 2024-09-20
splits CI build

kdschlosser:
Splits the CI build into separate files and removes the need to decompress an archive of the HTML output. This also adds building using Windows as well. 

----
📚 Documentation preview 📚: https://writethedocs-www--2229.org.readthedocs.build/


plaindocs (commented):
This looks great to me.

@mxsasha want to put eyes on this before we merge it? tl;dr, I opened #2226 to add a Windows CI build to help keep WTD buildable on Windows. And @kdschlosser suggested splitting the build up, and did so, as well as adding in Windows.

plaindocs:
And those three waiting jobs are I assume from master branch, and will be removed when we merge this in.

kdschlosser:
They should disappear when this gets merged. 

Like how everything now builds in parallel? a lot more efficient this way especially now that you have the Linux and the Windows builds which take some time to run.

kdschlosser:
It's also better having it split up because if you need to make a change to the CI and there is a problem that comes up it doesn't break everything. It only ends up breaking a small portion of it.

plaindocs:
Yep, love that last point. :-)

ericholscher (approved):
Looks like a solid change to me 👍

plaindocs:
ok, with the second :+1: I'll merge this in.

kdschlosser:
Thanks guys. I believe this is definitely a step in the right direction since the CI is getting larger. The ability to have the CI run the tests and builds in parallel is the way to go.

plaindocs:
For posterity, if you have specific rules mention in branch protection rules, you have to go update those after changing the CI names. :+1:

kdschlosser:
we kind of touched base on that a bit when you asked about those 3 rules and them disappearing. I said they "should" and if they didn't then you would need to go into the actions and make adjustments as needed. 

looks like you figured that out tho.

plaindocs:
Absolutely not pointing at you there ;-), I just read the first half and missed the second. All good.

kdschlosser:
no worries m8. This is a learning experience for us all, and that is the reason why we are doing this stuff. It's to learn and to share knowledge.

PR #2240, opened 2024-10-07
RTD updates

plaindocs:
----
📚 Documentation preview 📚: https://writethedocs-www--2240.org.readthedocs.build/

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -50,7 +50,7 @@ sys.path.append(os.getcwd())  # noqa
 # https://about.readthedocs.com/blog/2024/07/addons-by-default/
 
 # Define the canonical URL if you are using a custom domain on Read the Docs
-html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "www.writethedocs.org")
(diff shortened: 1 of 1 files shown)


mxsasha (approved):

plaindocs:
There is also config in the GH settings for that, so let's make sure that's the issue before committing ?

mxsasha:
> There is also config in the GH settings for that, so let's make sure that's the issue before committing ?

Don't think it is on closer look. I think we're missing https:// prefix - trying that now. Might cause issues with PR builds and link checks though, as links won't be available on www.writethedocs.org and maybe there will be misdirected absolute URLs, that should have the PR URL instead. But I do want to get the recap out, so if this fixes that part, it'll have to do for now.

plaindocs:
The content looks ok, so it *might* be just the linkchecker config

I'd rather disable that temporarily than revert the CI

mxsasha:
> The content looks ok, so it _might_ be just the linkchecker config

Revert is already dropped, see other comment :)

PR #2246, opened 2024-10-23
Sphinx etc version upgrades

plaindocs:
We don't have to merge this yet, and especially without testing, but I went down this rabbit hole because I wanted to upgrade ablog to 0.11.11 which removes the spurious parallelization error. 

That led to a few other changes and crashed into

```
Extension error (ablog):
Handler <function builder_inited at 0x7fd8867c5b40> for event 'builder-inited' threw an exception (exception: Found the path from `ablog.get_html_templates_path()` in the `templates_path` variable from `conf.py`. Doing so interferes with Ablog's ability to stay compatible with Sphinx themes that support it out of the box. Please remove `get_html_templates_path` from `templates_path` in your `conf.py` to resolve this.)
make: *** [Makefile:55: html] Error 2
```

Which probably looks like it'll take more significant work.

----
📚 Documentation preview 📚: https://writethedocs-www--2246.org.readthedocs.build/

### DIFF
--- a/docs/_templates/2022/generic.html
+++ b/docs/_templates/2022/generic.html
@@ -86,7 +86,7 @@
                 {% block body %}
                 {{ body }}
                 {% endblock body %}
-                {# {% include "postcard.html" %} #}
+                {# {% include "ablog/postcard.html" %} #}
         </div>
     </div>
   {% endif %}
--- a/docs/_templates/2023/generic.html
+++ b/docs/_templates/2023/generic.html
@@ -86,7 +86,7 @@
                 {% block body %}
                 {{ body }}
                 {% endblock body %}
-                {# {% include "postcard.html" %} #}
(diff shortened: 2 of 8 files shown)


plaindocs:
OK, maybe it's not actually so bad. Just a bunch of these to sort out.

```
WARNING: in rendering sponsorship/jobs.html: postcard.html is an old template path, that is no longer used by
ablog. Please use ablog/postcard.html instead.
WARNING: in rendering sponsorship/newsletter.html: postcard.html is an old template path, that is no longer used by
ablog. Please use ablog/postcard.html instead.
WARNING: in rendering sponsorship/slack.html: postcard.html is an old template path, that is no longer used by
```

plaindocs:
OK, so: ``alog/postcard.html`` provides information regarding the current post (when on a post page)

PR #2256, opened 2024-11-07
[Australia] - Tweaks to Day 2 schedule

swapnilogale:
----
📚 Documentation preview 📚: https://writethedocs-www--2256.org.readthedocs.build/


swapnilogale:
@plaindocs - tweaked Day 2 as we have less one speaker, and a Tales from the Trenches session instead of regular LTs.

plaindocs (approved):
That looks fine, I didn't checking the times in anger.

You've still got the dupe "title: Brick Docs Workshop" which I think you could just remove and leave the linked one?

PR #2271, opened 2024-12-18
Update Python from 3.8 to 3.9

nickspaargaren:
### What

- Updates the project used Python version from 3.8 to 3.9
- Why? 3.8 has become unsupported, see: https://devguide.python.org/versions/#supported-versions

### How to test
- Follow the "Installing the project" [instructions](https://github.com/writethedocs/www?tab=readme-ov-file#prerequisites-for-generating-the-docs-locally).
- Make sure the projects runs on Python 3.9
- Check out the Python 3.9 [deprecated section](https://docs.python.org/3/whatsnew/3.9.html#deprecated) for breaking changes 
- All GitHub Actions (pipelines) should pass

----
📚 Documentation preview 📚: https://writethedocs-www--2271.org.readthedocs.build/


requirements.txt
```diff
@@ -1,17 +1,18 @@
-Sphinx==4.5.0
+Sphinx>=5,<6
```
ericholscher: This is the only thing that scares me. I don't see anything obviously broken tho, and if the build doesn't have warnings we're probably fine.

ericholscher (commented):
Thanks for your work here!

If we're doing this work, I wonder if it makes sense to upgrade to 3.10 or 3.11, so we're a bit closer to the latest version? Might cause more cascading dependency upgrades though.

requirements.txt
plaindocs: Yeah, I'd be pro moving all the way up as well. I actually started looking at that over in https://github.com/writethedocs/www/pull/2246 and abandoned it though.

docs/about/learning-resources.rst
```diff
@@ -7,7 +7,6 @@ Learning Resources
    :hidden:
 
    /topics
-   /videos/index
```
ericholscher: This is still used in production, just not built locally. 

https://github.com/writethedocs/www/blob/14230d289fba7c5889b2d2bf252c269efa2c4477/docs/conf.py#L97-L104

plaindocs:
I took a quick look at the two broken links, and they seem pretty irrelevant to this PR, which does make me wonder why they aren't failing the main build:

```
 For the Links > Internal check, the following failures were found:

* At docs/_build/html/blog/archive/tag/index.html:75:

  internally linking to ../../../../, which does not exist

* At docs/_build/html/blog/archive/tag/index.html:140:

  internally linking to ../../../../, which does not exist
```

nickspaargaren:
> I took a quick look at the two broken links, and they seem pretty irrelevant to this PR, which does make me wonder why they aren't failing the main build:
> 
> 
> 
> ```
> 
>  For the Links > Internal check, the following failures were found:
> 
> 
> 
> * At docs/_build/html/blog/archive/tag/index.html:75:
> 
> 
> 
>   internally linking to ../../../../, which does not exist
> 
> 
> 
> * At docs/_build/html/blog/archive/tag/index.html:140:
> 
> 
> 
>   internally linking to ../../../../, which does not exist
> 
> ```

Yeah i'm seeing this as well. I was trying different pip modules/package versions to tackle this. No solution yet.. Maybe there are indeed wrong links?

plaindocs:
Yeah it's probably a difference in behaviour from the HTML output? 

This ugliness is L75:

```html
          <div class="breadcrumbs">
            
            
              <a href="../../../../">Home</a> &raquo;
                
            
          </div>
```

The href here is L140: 

```html
<p class="logo">
  <a href="../../../../">
    <img class="logo" src="../../../../_static/sticker-wtd-colors.png" alt="Logo" />
    
  </a>
</p>
```

Could track down where it's generated, and if there isn't an obvious fix, just ignore it? :-p 

https://github.com/gjtorikian/html-proofer?tab=readme-ov-file#ignoring-content

plaindocs:
If we want to make this pass I'm happy to exclude ` docs/_build/html/blog/archive/tag/index.html` manually in the htmlproofer call in CI and add an issue to look into it at another date.

nickspaargaren:
> If we want to make this pass I'm happy to exclude ` docs/_build/html/blog/archive/tag/index.html` manually in the htmlproofer call in CI and add an issue to look into it at another date.

Thanks for understanding! Let me look into it one more time, maybe i can come up with a fix. Otherwise let's ignore it.

plaindocs:
I wonder if diffing the ablog source between the versions would help? I don't *think* we've got custom ablog templates in our repo.

nickspaargaren:
@plaindocs After another deep dive, I have no idea were this is coming from. Do you want to exclude this file in the `htmlproofer` ignore list for me in this branch?

plaindocs:
> @plaindocs After another deep dive, I have no idea were this is coming from. Do you want to exclude this file in the `htmlproofer` ignore list for me in this branch?

Yeah, my gut feel is that the updated ablog is expecting a different setting somewhere which is generating those two links. I might have another poke another time, but it certainly doesn't feel important.

plaindocs (approved):
Thanks for being patient and pushing this over the line.

plaindocs:
:tada:

PR #2274, opened 2024-12-20
Remove sphinx-action from build step and build HTML directly

nickspaargaren:
### What
- When working on the python update I ran into a problem of python version mismatch. This PR fixes that.
- Others also ran into this problem and decided to build the HTML directly:
https://github.com/ammaraskar/sphinx-action/issues/43#issuecomment-1218439431

### How to test
- Pipeline should pass.
- HTML should build as normal.

### Notes
- This loses the [checks](https://github.com/marketplace/actions/sphinx-build-composite) function
- The [Github action](https://github.com/rickstaa/action-sphinx-composite) isn't updated in 4 years

----
📚 Documentation preview 📚: https://writethedocs-www--2274.org.readthedocs.build/

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -75,8 +75,6 @@ exclude_patterns = [
     'videos/prague/2018/tackling-technical-debt-in-the-docs-louise-fahey.rst',
 ]
 
-html4_writer = True
-
 # We use these *local* environment variables for private info like free ticket links
 
 cfp_variables = {}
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -75,8 +75,6 @@ def _load_data(self, env, data_source, encoding):
     'videos/prague/2018/tackling-technical-debt-in-the-docs-louise-fahey.rst',
 ]
 
-html4_writer = True
```
ericholscher: This seems unrelated? I'm :+1: on this change without this here.

nickspaargaren: Check! Pipeline was failing before i believe, that's why i ran into this. Resolved!

ericholscher (approved):

PR #2285, opened 2024-12-31
Add option to locally run HTMLProofer

nickspaargaren:
### What
This PR adds the option to locally run [HTMLProofer](https://github.com/gjtorikian/html-proofer). With the Python upgrade https://github.com/writethedocs/www/pull/2271 I ran into a problem with the html generation. This makes it easier to run the check without waiting for the Github actions to finish.

### How to test

1. Check out this branch
2. Open VSCode and (Re)open in container. Or use: `>Dev Containers: Reopen in Container`
3. Wait for the container build to finish.
4. Use `cd docs/` to navigate to the docs folder
5. Run `make html` to generate the html inside `docs/_build`
6. Use `cd ../` to navigate back to the root folder
7. Run the htmlproofer command used in https://github.com/writethedocs/www/blob/main/.github/workflows/ubuntu.yml
   - `htmlproofer --ignore-files "/404/,/2013/,/2014/,/2015/,/2016/,/2017/,/search\/index.html/" --allow-hash-href=true --enforce-https=false --ignore-missing-alt=true --disable-external=true docs/_build/html`
8. Make sure the command can be run and checks for html warnings.

### TODO

- [x] Make sure Github actions pass with new `ruby/setup-ruby` action.

----
📚 Documentation preview 📚: https://writethedocs-www--2285.org.readthedocs.build/


plaindocs:
I'd be open to moving the htmlproofer link check out into external config or script to make it easier to run both locally and in CI.

The issue I ran into when running locally was that there are a lot of links like "/site-map/" which fail locally, and we'd have to figure out a way to address those. Or run it against a locally running server. :man_shrugging: Also a load of options which are valid in CI aren't locally (also probably a different version thing).

This part of the tool chain is really not a joy.

nickspaargaren:
> I'd be open to moving the htmlproofer link check out into external config or script to make it easier to run both locally and in CI.
> 
> The issue I ran into when running locally was that there are a lot of links like "/site-map/" which fail locally, and we'd have to figure out a way to address those. Or run it against a locally running server. 🤷‍♂️ Also a load of options which are valid in CI aren't locally (also probably a different version thing).
> 
> This part of the tool chain is really not a joy.

I saw this also locally. Maybe this is something we can fix later if the sphinx-ubuntu passes?

My secondary goal is to keep the `htmlproofer` updated with dependabot as well. At first i thought it was a bug that was fixed in a newer version of `htmlproofer`.

.github/workflows/ubuntu.yml
```diff
@@ -56,6 +53,12 @@ jobs:
       with:
         name: html
         path: html.tar.gz
+
+    - name: Set up Ruby
+    - uses: ruby/setup-ruby@v1
```
plaindocs: Looks like

```suggestion
      uses: ruby/setup-ruby@v1
```

.github/workflows/ubuntu.yml
```diff
@@ -56,6 +53,15 @@ jobs:
...
         path: html.tar.gz
+
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: '3.4'
+        bundler-cache: true
+        
+      - name: Install Ruby dependencies
+      run: bundle install
```
plaindocs:
```suggestion
      - name: Install Ruby dependencies
        run: bundle install
```

.github/workflows/ubuntu.yml
```diff
@@ -56,6 +53,15 @@ jobs:
...
         path: html.tar.gz
+
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: '3.4'
+        bundler-cache: true
+        
+      - name: Install Ruby dependencies
+        run: bundle install
```
plaindocs: Probably :p - sorry for the mess

```suggestion
    - name: Install Ruby dependencies
      run: bundle install
```

plaindocs (commented):
I've not tested the local side, but the remote looks like it's working.

Do you want to check in a broken link to make sure it crashes out, then we can revert and merge?

nickspaargaren:
> I've not tested the local side, but the remote looks like it's working.
> 
> Do you want to check in a broken link to make sure it crashes out, then we can revert and merge?

Sure thing, let me know if you've seen the [failed pipeline](https://github.com/writethedocs/www/actions/runs/12875435084/job/35896720960?pr=2285) so I can revert the temporary commit!

plaindocs (approved):
Lovely. :+1:

nickspaargaren:
> Lovely. 👍

Temporary commit has been removed, ready to merge!

PR #2292, opened 2025-01-15
Don't blockquote the sponsor list

ericholscher:
----
📚 Documentation preview 📚: https://writethedocs-www--2292.org.readthedocs.build/


plaindocs (approved):
I wonder if there is a flag to remove the white space with the templating, and keeping the indent. Nevermind.

PR #2312, opened 2025-02-11
Add banner to top of the docs pages with announcements

ericholscher:
Screenshot 2025-02-11 at 10 07 46 AM

----
📚 Documentation preview 📚: https://writethedocs-www--2312.org.readthedocs.build/

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -1,6 +1,11 @@
 {%- extends "!layout.html" %}
 
 {%- block document %}
+  <div class="announcement-banner-wrapper">
+    <div class="announcement-banner">
+      <p>{{ announcement_message | safe }}</p>
+    </div>
+  </div>
   <div class="documentwrapper">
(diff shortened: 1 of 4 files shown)


docs/_templates/layout.html
```diff
@@ -1,6 +1,11 @@
 {%- extends "!layout.html" %}
 
 {%- block document %}
+  <div class="announcement-banner-wrapper">
+    <div class="announcement-banner">
+      <p>{{ announcement_message | safe }}</p>
+    </div>
+  </div>
```
plaindocs, line 10: I wonder if we could use something like 

```
.. postlist:: 1
```

perhaps with a check on whether Portland is active using `flagpostconf: False` to just show the last news post?

ericholscher: It's kind of a pain to bring in content like that, so I think driving it from the config is probably easiest for now?

plaindocs (approved):
:shipit:

PR #2321, opened 2025-02-21
New speaker pic

plaindocs:
----
📚 Documentation preview 📚: https://writethedocs-www--2321.org.readthedocs.build/


ericholscher (approved):
I'd like to keep these the same dimensions if possible. I think there's a command to run on Mac to resize them noted in the speaker page, but there should be something similar on Linux.

plaindocs:
Can we just set a CI job to open a PR on daily basis?

plaindocs:
And where do we document what size we want them to be?

PR #2325, opened 2025-02-25
Add virtual tickets

ericholscher:
This builds the infra on the website for virtual tickets.

Next steps:

* Tickets have been created on tito, but need to be unhidden before this goes live.
* Need to write a blog post announcing them.

----
📚 Documentation preview 📚: https://writethedocs-www--2325.org.readthedocs.build/

### DIFF
--- a/docs/_data/schema-config.yaml
+++ b/docs/_data/schema-config.yaml
@@ -92,6 +92,7 @@ ticket-details:
   corporate: include('price-type')
   independent: include('price-type')
   student: include('price-type')
+  virtual: include('price-type')
 
 price-type:
   price: str()
--- a/docs/_data/schema-config.yaml
+++ b/docs/_data/schema-config.yaml
@@ -92,7 +92,7 @@ ticket-details:
   corporate: include('price-type')
   independent: include('price-type')
   student: include('price-type')
-  virtual: include('price-type')
+  virtual: include('price-type', required=False)
 
 price-type:
   price: str()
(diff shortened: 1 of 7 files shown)


docs/conf/portland/2025/virtual.md
```diff
@@ -0,0 +1,39 @@
...
+
+We'll be hosting the virtual conference on [Venueless](https://venueless.org), 
+a platform that combines video streaming with chat and networking features.
+
+## What's Included
+
+Virtual attendance includes:
+
+* Live streamed talks from the main stage
+* Real-time chat with all attendees
```
janine-c: I don't think in-person attendees are going to be online?
```suggestion
* Real-time chat with other virtual attendees
```

docs/conf/portland/2025/virtual.md
```diff
@@ -0,0 +1,39 @@
...
+* Virtual networking spaces  
+* Access to talk recordings after the event
```
janine-c: I'm not sure I'd include this? You also get access to recordings if you have no ticket at all. I wouldn't want to make people think that they have to buy a ticket of some sort to get access.

docs/conf/portland/2025/virtual.md
plaindocs: That's actually an interesting point. If we keep the chat in something like slack its more likely that folks will be there from the in person conference.

docs/_data/schema-config.yaml
```diff
@@ -92,6 +92,7 @@ ticket-details:
   corporate: include('price-type')
   independent: include('price-type')
   student: include('price-type')
+  virtual: include('price-type')
```
plaindocs: Make this optional to prevent the YAML validation error?

plaindocs (commented):
Seems good.

PR #2379, opened 2025-05-30
N/A 2

Thoughtscript:


plaindocs:
Hey @Thoughtscript this is pretty cool. I especially like how you've reused the make script.

docs/Makefile
```diff
@@ -59,6 +59,9 @@ html:
 livehtml:
 	sphinx-autobuild --port 8888 -j auto --watch _data -b dirhtml $(ALLSPHINXOPTS) $(BUILDDIR)/html
 
+dockerhtml:
```
plaindocs, line 63: Worth commenting here that this is for use inside the Docker image, not externally to run it?

docs/make.bat
```diff
@@ -18,6 +18,7 @@ if "%1" == "help" (
 	echo.Please use `make ^<target^>` where ^<target^> is one of
 	echo.  html       to make standalone HTML files
 	echo.  livehtml   to make live HTML files to view the docs on a local web server
+	echo.  dockerhtml to make live HTML files and view through a local docker container
```
plaindocs, line 21: Definitely worth clarifying here.

run.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+cd docs && make dockerhtml &
+
+wait
```
plaindocs, line 5: Don't suppose we can fold this inside the Dockerfile and drop the file?

README.md
```diff
@@ -108,6 +108,17 @@ If you have verified this command on Windows, we invite you to submit a PR to in
...
+### Using Docker
+
+A barebones [dockerfile](./dockerfile) is supplied to run the site within a local Docker Container through [Docker Desktop](https://docs.docker.com/desktop/) - a simple, free, way to easily setup Docker and Python without leaving beyond installations, modifying your underlying Operating System, and changing Environment Variables.
+
+1. Make sure Docker Desktop is running and started
+1. Build the Docker Image from the root directory with the command: `docker build .`
+1. Find the created Docker Image ID: `docker images` - something like: `sha256:123...`
+1. Run the Docker Image Container using that Image ID: `docker run -p 8888:8888 sha256:123...`
+1. Access the live site on <http://localhost:8888> through your web browser
+1. Both the Docker Container and Image will be present in Docker Desktop
```
plaindocs: Would it make sense to tag the dockerfile as `wtd` on creation, and then run the tag? Probably not, you still need the image sha.

Is there no way to just say "run the image build from the current dir" without going around looking for the ID?

plaindocs (commented):
This is great. I've left a few suggestions or queries.

docs/Makefile
Thoughtscript: Used solely to bind sphinx to the Docker Container - to do so requires deliberate --host "0.0.0.0" - will add a comment

docs/make.bat
Thoughtscript: Used solely to bind sphinx to the Docker Container - to do so requires deliberate --host "0.0.0.0" - will add a comment

run.sh
Thoughtscript: I still think this is the best way to handle this since it's by far the easiest. It also avoids creating permissions and allows for easier extensibility. Since you clearly disagree, I'll just go ahead and CLOSE this ticket.

PR #2381, opened 2025-06-04
Add June 2025 newsletter

CollierCZ:
----
📚 Documentation preview 📚: https://writethedocs-www--2381.org.readthedocs.build/

### DIFF
--- a/docs/blog/newsletter-june-2025.rst
+++ b/docs/blog/newsletter-june-2025.rst
@@ -38,7 +38,7 @@ See more Write the Docs resources about `release notes </topics/#release-notes>`
 Automating screenshot capture or translation
 --------------------------------------------
 
-The quantity of screenshots in documentation has been an ongoing issue for many years. The general consensus has been to minimize the number of screenshots — be selective about what to use. For an example, see the 2022 newsletter article `When to create screenshots </blog/newsletter-december-2022/#when-to-create-screenshots>`__. In the past, the concerns usually focused on accessibility, ROI, and maintenance issues. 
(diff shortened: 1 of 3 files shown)


docs/blog/newsletter-june-2025.rst
```diff
@@ -0,0 +1,108 @@
...
+
+When done well, release notes showcase progress, reduce support issues, and reinforce a user-first mindset.
+
+See more Write the Docs resources about `release notes </topics/#release-notes>`__.
+
+--------------------------------------------
+Automating screenshot capture or translation
+--------------------------------------------
+
+The quantity of screenshots in documentation has been an ongoing issue for many years. The general consensus has been to minimize the number of screenshots — be selective about what to use. For an example, see the 2022 newsletter article `When to create screenshots </blog/newsletter-december-2022/#when-to-create-screenshots>`__. In the past, the concerns usually focused on accessibility, ROI, and maintenance issues.
```
hillaryfraley: Do we need to define ROI here at first use? maybe that's overkill.

hillaryfraley (approved):

docs/blog/newsletter-june-2025.rst
CollierCZ: I was thinking about that too so I guess if we both were, it makes sense to include.

```suggestion
The quantity of screenshots in documentation has been an ongoing issue for many years. The general consensus has been to minimize the number of screenshots — be selective about what to use. For an example, see the 2022 newsletter article `When to create screenshots </blog/newsletter-december-2022/#when-to-create-screenshots>`__. In the past, the concerns usually focused on accessibility, return on investment (ROI), and maintenance issues. 
```

PR #2420, opened 2025-08-11
Fix remaining Pretalx scripts

mxsasha:
@plaindocs perhaps we also want to switch out twitter for bluesky in the questions output, I left that as is now

----
📚 Documentation preview 📚: https://writethedocs-www--2420.org.readthedocs.build/

### DIFF
--- a/docs/_scripts/pretalx2mc-info.py
+++ b/docs/_scripts/pretalx2mc-info.py
@@ -10,7 +10,7 @@ import sys
 sys.path.insert(1, '../_ext')
 
 import os
-from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


plaindocs:
I think so yes, although perhaps @katiejano still wants both?

katiejano:
In Portland the decision was to officially stop using Twitter so yes, let's remove it and add bluesky.

katiejano:
@plaindocs will you give me access to pretalx?

plaindocs (approved):
Looks good to me

plaindocs:
@mxsasha merging this, we can fix the Twitter / Bluesky thing after.

PR #2423, opened 2025-08-19
Move past conferences off front page

mxsasha:
----
📚 Documentation preview 📚: https://writethedocs-www--2423.org.readthedocs.build/

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -218,8 +218,8 @@ global_sponsors = yaml.safe_load("""
 # Dynamic announcement message
 announcement_message = None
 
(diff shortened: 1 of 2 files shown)


docs/include/conf/current.rst
```diff
@@ -1,7 +1,5 @@
 2025 Conferences
```
ericholscher: We normally keep all the conferences for the year on the front page, but we could add a banner or highlight Berlin first in the list?

There's some code in the conf.py file from the banner we put in Portland that I'm guessing should be easy to reuse.

mxsasha: Sounds good. I kept it on the front page, but broke the list in two - making the order not chronological looked odd too. Banner added.

plaindocs (approved):

PR #2451, opened 2025-10-09
Update refund policy for in-person->virtual

mxsasha:
We've had this question a few times, and this is the way I've dealt with them so far. I think also here the two week limit makes sense.

----
📚 Documentation preview 📚: https://writethedocs-www--2451.org.readthedocs.build/


docs/conf/berlin/2025/tickets.rst
```diff
@@ -44,6 +44,8 @@ Refund Policy
 -------------
 
 Refunds are offered with a 10% processing fee, up to 2 weeks before the conference.
+This includes changing your in-person ticket to virtual, where we will refund the price difference,
+minus the processing fee.
```
plaindocs: I think this is kinda confusing. Are we also saying that after 2 weeks before the conference you can't change to virtual with a 10% fee? 

So how about something like. 

- Refunds are offered with a 10% processing fee, up to 2 weeks before the conference.
- Changing an in-person ticket to a virtual ticket also incurs a 10% processing fee.

plaindocs: And if we are saying that, add a note to the second bullet point about the timeline?

mxsasha: I think we should also apply the two week limit to virtual ticket conversion.

plaindocs:
@mxsasha we should update this for 2026 and merge I think

mxsasha:
> @mxsasha we should update this for 2026 and merge I think

Done.

plaindocs (approved):

PR #2473, opened 2025-11-12
Portland 2026: Copy Berlin approach to virtual

ericholscher:
This is a nicer way of displaying this data,
so let's use it here.

----
📚 Documentation preview 📚: https://writethedocs-www--2473.org.readthedocs.build/


plaindocs (approved):
Looks good, I didn't test the output.

PR #2480, opened 2025-11-24
Delete Dependabot configuration

ericholscher:
We never use these, so let's stop spamming ourselves. We can upgrade when we're ready. 

----
📚 Documentation preview 📚: https://writethedocs-www--2480.org.readthedocs.build/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,16 +1,1 @@
-# Set update schedule for GitHub Actions
 
-version: 2
-updates:
-  - package-ecosystem: "pip"
-    directory: "/"
-    schedule:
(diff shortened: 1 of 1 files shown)


plaindocs (approved):
I think the issue is that we've got a few things we can't easily update. 

We could just bump the check to six months or something, so we're not spammed, or pin/ignore the things that we cant update

Or just reup when we have updated :-p

ericholscher:
Yea, I think we just re-enable it when we can.

plaindocs:
Not entirely sure why we're still getting https://github.com/writethedocs/www/pull/2488

There is a separate GH repo setting, but it's disabled there.

PR #2510, opened 2026-01-28
update buttons sponsors

ericholscher:
- **Update RTD install guide**
- **Update buttons and sponsors**

----
📚 Documentation preview 📚: https://writethedocs-www--2510.org.readthedocs.build/


docs/_data/portland-2026-config.yaml
```diff
@@ -219,6 +214,18 @@ sponsors:
   first:
   in_kind:
 
+
+photos:
+  default:
```
diggerlilly, line 219: @ericholscher are these meant to be updated to `/2026` ?

If so, there's no `grants.jpg` there.

ericholscher: No, these just moved down. We're reusing the same as last year.

PR #2529, opened 2026-02-06
update Atlassian CLI principles blog post url

bergzand:
The old url is no longer valid, the Atlassian blog has moved. This updates the url to the new blog location.

I suspect this was not caught by any automated checks as the old url was redirected to a generic main page, and thus appears as working to the check.

----
📚 Documentation preview 📚: https://writethedocs-www--2529.org.readthedocs.build/


plaindocs (approved):
Thanks!

PR #2569, opened 2026-03-07
Logo fix

plaindocs:
----
📚 Documentation preview 📚: https://writethedocs-www--2569.org.readthedocs.build/


ericholscher (approved):
I usually swap it as a `-new` logo, so the historical aren't effected, but probably fine to just swap it.

PR #2577, opened 2026-03-16
Update Berlin post date from March 5 to March 16, maybe fix mailchimp

mxsasha:
Mailchimp claims this may cause it to be sent out at last.

----
📚 Documentation preview 📚: https://writethedocs-www--2577.org.readthedocs.build/


plaindocs (approved):
Believe it when I see it ;-)

PR #2589, opened 2026-03-25
bullet fixes applied to updated CSS

goatlady:
NEW PR wth CSS changes applied to updated main to avoid stupid conflicts:

Removed custom bullet marker; changed bullet color with ::marker CSS; restored sensible defaults for padding and indentation.

Bulleted list indentation is now consistent throughout the body copy regardless of content length. Second level bullets now have a different bullet style (HTML default, but custom color) - this is maybe better for accessibility? But could be changed to the same style if we think it doesn't look good.

image

----
📚 Documentation preview 📚: https://writethedocs-www--2589.org.readthedocs.build/


janine-c (approved):
Stunning! Thanks so much for fixing what I couldn't figure out! 😍

PR #2606, opened 2026-04-07
Add button-link directive to replace inline HTML buttons

ericholscher:
## Summary
- Adds a `button-link` Sphinx directive (`docs/_ext/button.py`) that replaces verbose inline HTML table buttons with a clean syntax:
  ````
  ```{button-link} https://example.com
  Button text
  ```
  ````
- Replaces all 5 inline HTML button instances across 3 Portland 2026 news posts
- Renders identical HTML output to the previous inline approach

## Test plan
- [ ] Verify RTD preview build renders buttons correctly on the 3 news pages
- [ ] Check button styling matches previous inline HTML buttons

----
📚 Documentation preview 📚: https://writethedocs-www--2606.org.readthedocs.build/

### DIFF
--- a/docs/_ext/button.py
+++ b/docs/_ext/button.py
@@ -1,0 +1,38 @@
+from docutils import nodes
+from docutils.parsers import rst
+
+
+class ButtonLink(rst.Directive):
+    """A simple button directive that renders a styled call-to-action button.
+
+    Usage in myst markdown::
+
+        ```{button-link} https://example.com
+        Button text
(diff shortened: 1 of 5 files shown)


plaindocs (approved):
Nice!

plaindocs:
Might wanna check with @goatlady that this works on mobile before we ship

ericholscher:
Should be the exact same HTML output.

PR #2605, opened 2026-04-06
Remove video archive and link to YouTube channel

ericholscher:
## Summary

Supersedes #2482 — rebased onto current `main`, resolved conflicts, and cleaned up lingering references.

- Remove the video Sphinx extension (`videos.py`) and related `conf.py` logic
- Update speaker YAML files (2015–2018) to link directly to YouTube videos
- Update newsletter/blog posts and `topics.rst` video links to point to YouTube channel
- Update `videos/index.rst` to direct users to the YouTube channel
- Remove `BUILD_VIDEOS` env var from all CI workflow files
- Remove outdated video instructions from `README.md`, `AGENTS.md`, `CLAUDE.md`, and `CONFERENCE-SITES.md`
- Clean up redundant `conf.py` exclude pattern

## Test plan

- [x] `make html` builds with no new warnings
- [x] YAML schema validation passes (`validate-yaml.sh`)
- [x] Spellcheck clean for changed files
- [ ] CI passes on this branch

https://claude.ai/code/session_01D7qkyurXYZmJ7zBsXSAn6v

----
📚 Documentation preview 📚: https://writethedocs-www--2605.org.readthedocs.build/

PR #2608, opened 2026-04-07
Remove video RST source files and unused video templates

ericholscher:
## Summary

- Delete 343 video RST source files under `docs/videos/` (individual talk pages)
- Delete 3 unused video templates (`video-detail.html`, `video-listing.html`, `meetup-listing.html`)

These files used the `datatemplate-video` directive from the now-deleted `videos.py` extension and were already excluded from builds via the `videos/**` pattern in `conf.py`. This is a pure cleanup — no functional change.

Follow-up to #2605 which removed the video archive system. Split into a separate PR because the deletion of 346 files exceeds GitHub's 300-file diff API limit, which breaks the Vale CI action's reviewdog integration.

## Test plan

- [ ] CI passes (Vale will fail due to 300-file limit, but that's expected for delete-only PRs — no content to lint)
- [x] Files are already excluded from builds, so no build impact

https://claude.ai/code/session_01D7qkyurXYZmJ7zBsXSAn6v

----
📚 Documentation preview 📚: https://writethedocs-www--2608.org.readthedocs.build/


plaindocs:
oh yeah, this was a nonsense build failure from vale

PR #2612, opened 2026-04-07
Update join link

plaindocs:
We should move this to an environment variable on RTD maybe? Seems kinda silly to open PRs every month or so for it.

----
📚 Documentation preview 📚: https://writethedocs-www--2612.org.readthedocs.build/


ericholscher (approved):
Yea... or find a better solution for the overall problem.

PR #2626, opened 2026-04-13
Update devcontainer config

mxsasha:
Updated to work with our current uv setup, and supports local overrides from ~/.devcontainer/docker-compose.local.yml.

I want to use more devcontainers for isolation, and this allows me to still attach it to my ssh key used for git, along with other local tweaks. It still allows local htmlproofer (#2285), the ruby isntall is a bit weird because of broken ruby images.

From the file history, I think nobody else is actively using this? If anyone is, we should make sure their workflow still works.

----
📚 Documentation preview 📚: https://writethedocs-www--2626.org.readthedocs.build/

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,8 @@
+FROM mcr.microsoft.com/devcontainers/python:1-3.12-bookworm
+
+# Remove broken yarn apt source whose GPG key is expired (devcontainers/images#1752)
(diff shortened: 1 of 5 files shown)


ericholscher:
Seems @goatlady is using them too? https://github.com/writethedocs/www/pull/2620

I've played with them, should probably start trying them again.

ericholscher (approved):

PR #2635, opened 2026-04-14
Fix per-conference CoC pages broken by heading rename in #2633

ericholscher:
## Summary

PR #2633 renamed the team section headings in `docs/code-of-conduct.rst`:

- `Portland conference Code of Conduct Team` → `Portland Code of Conduct Team`
- `Berlin conference Code of Conduct Team` → `Berlin Code of Conduct Team`
- `Australia Code-of-Conduct Team` → `Australia Code of Conduct Team`

Eight per-conference CoC pages `.. include::` that file with `:end-before:` pointing at the old heading text. Sphinx now raises CRITICAL `Text not found` errors on every one of those pages, and the rendered CoC content is replaced by a system-message stub — so the conference CoC pages are effectively broken on the live site.

This PR updates the `:end-before:` markers to match the new heading text:

- `docs/conf/portland/{2024,2025,2026}/code-of-conduct.*`
- `docs/conf/berlin/{2025,2026}/code-of-conduct.*`
- `docs/conf/australia/{2024,2025}/code-of-conduct.rst`
- `docs/conf/atlantic/2024/code-of-conduct.rst`

## Why didn't CI catch this?

The Ubuntu Build job runs `make html` without `-W`, so docutils `CRITICAL` system messages from the `include` directive are emitted to stderr but don't fail the build — Sphinx exits 0 and publishes a page containing a `system_message` node in place of the included content. The `check_unparsed_markup.py` post-check only scans rendered HTML for literal directive syntax, which doesn't match the stub Sphinx emits here, so it passes too.

Worth considering as a follow-up: either run the build with `-W --keep-going` in CI, or extend `check_unparsed_markup.py` to flag `system_message` / `Problem with ... directive` text in the output.

## Test plan

- [x] Local `sphinx-build` — no more CRITICAL errors on the CoC pages
- [ ] CI build passes
- [ ] Spot-check a rendered conference CoC page (e.g. `/conf/portland/2026/code-of-conduct/`) to confirm the main CoC content renders above the Portland team section

https://claude.ai/code/session_014xQW73RQhCDgmV6xMq6ihZ

----
📚 Documentation preview 📚: https://writethedocs-www--2635.org.readthedocs.build/

PR #2636, opened 2026-04-14
Fail Sphinx builds on warnings (CI + RTD), fix pre-existing warnings

ericholscher:
## Summary

Turn the Sphinx build into a warnings-as-errors build in both CI and Read the Docs, then fix every pre-existing warning so it actually passes. This is a follow-up to #2635 — that PR fixes the broken Code of Conduct page; this PR makes sure the same class of failure (silent `CRITICAL` from docutils, silent `WARNING` from Sphinx) can't reach production again.

**Stacked on #2635** — this branch includes the CoC-include fix commit, and the fail-on-warnings enforcement depends on it to pass. Merge #2635 first.

## Build plumbing

- **`docs/Makefile`** — add `-W --keep-going` to the `html` target only (via a target-specific variable). `livehtml` and `dockerhtml` are left alone so local authoring isn't disrupted by transient warnings.
- **`.github/workflows/ubuntu.yml`** — no change needed; the Ubuntu job already runs `cd docs && uv run make html`, so it inherits the flag from the Makefile.
- **`.readthedocs.yaml`** — set `sphinx.fail_on_warning: true` so the RTD build also fails on warnings.

## Content fixes (11 pre-existing issues)

Running `sphinx-build -W --keep-going` on `main` (plus the #2635 CoC fix) surfaced 11 warnings — 3 docutils `ERROR`s, 1 Sphinx `WARNING`, and 7 duplicate-label warnings. Each is addressed here.

**Table-width mismatches (3 × docutils ERROR):**
- `docs/surveys/salary-survey/2023.rst:1289` — `:widths: 55 10 10 10 15` on a 3-column table → `60 15 25`
- `docs/surveys/salary-survey/2024.rst:2046` — `:widths: 55 10 10 10 15` on a 4-column table → `55 15 15 15`
- `docs/surveys/salary-survey/2024.rst:3219` — `:widths: 70 15 15` on a 4-column table → `55 15 15 15`

**Orphan page (1 × Sphinx `toc.not_included`):**
- `docs/videos/index.rst` — page is a stub redirecting to YouTube and isn't in any toctree. Marked `:orphan:`.

**Duplicate hyperlink targets (7 × Sphinx `duplicate label`):**

Portland 2026 and Berlin 2026 each include the same shared jinja-template RST content:
- `docs/code-of-conduct.rst` → via `conf/{portland,berlin}/2026/code-of-conduct.md`
- `docs/include/conf/portland/speaking-tips-2025.rst` → via `conf/{portland,berlin}/2026/speaker-info.md`
- `docs/include/conf/virtual/talk-recording-guide.rst` → via `conf/{portland,berlin}/2026/talk-recording-guidelines.md`

When the hosting page is a MyST markdown file, ```` ```{eval-rst}\n{% include ... %}\n``` ```` ends up parsing the shared content inline in each page, and every section heading inside it registers a global hyperlink target in Sphinx's `std` domain. Two conferences = duplicate targets = warning.

These sections are never cross-referenced — they're purely visual headings — so the duplicates are harmless. Options considered:

1. **Rename headings per-conference** — would require forking the shared file or string-substituting in jinja, losing the point of sharing.
2. **Convert the section headings to `.. rubric::`** — would drop the page structure and change how users see the page.
3. **Blanket `suppress_warnings = ['ref']`** — the warning isn't emitted with `type='ref'` in Sphinx 8.2 (see `sphinx/domains/std/__init__.py:959`) so this doesn't work, and even if it did it would be too broad.
4. **Targeted logging filter (chosen)** — `docs/_ext/warning_filter.py` adds a `logging.Filter` to the `sphinx` logger that drops `duplicate label <name>, other instance in ...` warnings only for a small hard-coded allow-list of labels from the shared includes (7 entries). Any new accidental duplicate still fires a warning and fails the build.

Registered as an extension in `docs/conf.py`.

## Why didn't this break before?

Same reason #2635 didn't get caught: the default `make html` exits 0 regardless of how many `WARNING`/`CRITICAL` lines it prints. Without `-W`, Sphinx silently publishes broken pages. This PR closes that gap.

## Test plan

- [x] Local: `cd docs && uv run make html` succeeds with `-W --keep-going` in the invocation line, 0 warnings
- [x] Local: same for `sphinx-build -b dirhtml -W --keep-going` (matches RTD builder)
- [ ] CI: Ubuntu Build job passes
- [ ] RTD: build passes with `fail_on_warning: true`
- [ ] Spot-check: `/surveys/salary-survey/2023/`, `/surveys/salary-survey/2024/` render the currency tables correctly
- [ ] Spot-check: `/conf/portland/2026/talk-recording-guidelines/` and `/conf/berlin/2026/talk-recording-guidelines/` both render section headings

https://claude.ai/code/session_014xQW73RQhCDgmV6xMq6ihZ

----
📚 Documentation preview 📚: https://writethedocs-www--2636.org.readthedocs.build/


plaindocs (approved):
enthusiastic hell yeah from me

mxsasha (approved):

PR #2641, opened 2026-04-16
Add explicit start times to Portland 2026 schedule

ericholscher:
## Summary

- Add explicit `time` values to all talks, lunch breaks, lightning talks, roundtable discussions, reception, and conference closing in the Portland 2026 schedule
- Prevents cascading time bugs when an upstream duration is adjusted — major events are now pinned to their intended start times
- Short breaks, Q&A, and zero-duration markers remain relative since they naturally flow between pinned events

## Test plan

- [ ] Verify `uv run make html` builds without errors
- [ ] Verify schedule page renders with correct times for all events
- [ ] Verify YAML validation passes (`yamale -s docs/_data/schema-schedule.yaml docs/_data/portland-2026-schedule.yaml`)

https://claude.ai/code/session_01YRAWfBR3SELJnWEaSCfHkY

----
📚 Documentation preview 📚: https://writethedocs-www--2641.org.readthedocs.build/

### DIFF
--- a/docs/_data/portland-2026-schedule.yaml
+++ b/docs/_data/portland-2026-schedule.yaml
@@ -18,7 +18,8 @@ writing_day:
   - time: '8:00'
     duration: '1:00'
     title: Doors open
-  - duration: '0:30'
+  - time: '9:00'
+    duration: '0:30'
     title: Morning introduction - Introduction to Sprints & Project Introductions
   - duration: '0:00'
     title: 'Projects team up and start <a href="../writing-day">writing</a>'
@@ -39,7 +40,8 @@ writing_day:
     title: Projects team up and start writing
   - duration: '0:00'
     title: New to Git and GitHub workshop
-  - duration: '2:30'
+  - time: '14:00'
+    duration: '2:30'
     title: '<a href="../writing-day/#roundtable-discussions">Roundtable Discussions</a>'
   - time: '16:30'
     duration: '0:00'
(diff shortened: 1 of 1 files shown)


plaindocs (commented):
I'm not sure this is the way. Anchoring lunch breaks, sure, but having the talks flow is why we did this in the first place.

I can't think how we can easily test it though

ericholscher:
Yea, I think the big thing is that we're adding a lot more "one off" things to the schedule this year which has caused some weirdness. I do think once we lock in the timing (eg. go to print with schedules), we should hardcode things more so that they don't get messed up. We've had a couple schedule bugs this year because of relative time offsets being hard to reason about.

plaindocs:
Is it worth writing code to 'freeze' it? Like transform the relative times to absolute when we're happy with it? But we'll inevitably want to make changes after :D

plaindocs:
I'll try to give this a proper look later today

docs/_data/portland-2026-schedule.yaml
plaindocs, line 77: This clashes with the Q&A

plaindocs: Both days

plaindocs (approved):
I think this is ok. I've eye balled it as much as I can.

ericholscher:
I'll probably hold off on this given we're just around the corner from the event, but still think it's a neat idea.

plaindocs:
Yeah, agreed, let's revisit after

PR #2701, opened 2026-06-01
conf: add Event structured data for upcoming conferences

ericholscher:
A conference landing page is a search result for an event, but search engines had no way to know that — no dates, no venue, no "this is happening on these days." `schema.org/Event` JSON-LD is what lets Google and Bing render event rich results (date, location, organizer) and surface the conference in their event experiences, which is where a lot of "documentation conference near me" discovery happens.

All the data already lives in each config; the only thing missing was machine-readable dates, so the dates are parsed from the existing `date.short` field. The markup is emitted **only for conferences that haven't happened yet** — search engines ignore (and can flag) structured data for past events, and a list of expired events isn't what we want to advertise.

It renders on the conference landing page only (one event per URL), degrades gracefully when a field like the venue is still `TBC`, and uses absolute URLs so the markup validates.

---
*Generated by AI*

---
_Generated by [Claude Code](https://claude.ai/code/session_01UNSNa3pskWPPFBoDB6o8Ax)_

----
📚 Documentation preview 📚: https://writethedocs-www--2701.org.readthedocs.build/

### DIFF
--- a/docs/_data/schema-config.yaml
+++ b/docs/_data/schema-config.yaml
@@ -119,6 +119,8 @@ date-details:
   short: str()
   tickets_live: str()
   month: str()
+  start: str(required=False)  # ISO 8601 start date, e.g. 2026-09-06
+  end: str(required=False)    # ISO 8601 end date, e.g. 2026-09-08
   conference: include('day-details', required=False)
   day_one: include('day-details', required=False)
   day_two: include('day-details', required=False)
--- a/docs/_data/berlin-2026-config.yaml
+++ b/docs/_data/berlin-2026-config.yaml
@@ -110,6 +110,8 @@ shirts:
 date: # how do we handle these? human readable would be nice.
   main: "**September 6-8, 2026**"
   short: Sep 6-8, 2026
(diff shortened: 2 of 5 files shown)


docs/_data/schema-config.yaml
```diff
@@ -119,6 +119,8 @@ date-details:
   short: str()
   tickets_live: str()
   month: str()
+  start: str(required=False)  # ISO 8601 start date, e.g. 2026-09-06
```
plaindocs, line 129: If we're sure we're not editing old configs, we could stop linting them, and make a bunch of fields mandatory.

plaindocs (approved):
Seems like a good idea. Didn't test the output.

docs/_data/schema-config.yaml
ericholscher: Yea that's a great idea.

ericholscher: https://github.com/writethedocs/www/pull/2702

plaindocs (approved):

PR #2699, opened 2026-05-31
[australia 2026]: added landing page content for AU 2026

swapnilogale:
----
📚 Documentation preview 📚: https://writethedocs-www--2699.org.readthedocs.build/

PR #2702, opened 2026-06-02
config: only lint 2026+ configs and require more fields

ericholscher:
Picks up the suggestion from https://github.com/writethedocs/www/pull/2701#discussion_r3335870833: now that we're not editing the older conference configs, stop linting them and make a bunch more fields mandatory on current configs.

The config validation loop now skips anything older than 2026, so frozen conferences aren't validated. With those out of the way:

- **Previously-optional fields become required** — but only ones actually used in the site templates, content pages, includes, or extensions: `color`, `rgb`, `rgb_contrast`, `time_format`, `buttons`, `sponsorship`, `sponsors`, `cfp`, `grants`, `shirts`, `unconf`, `volunteer`, the `date.conference` block, venue details, and the core status flags.
- **Unused field removed** from the configs and schema: `about.location` (not referenced anywhere).
- **Stays compatible with the open Australia 2026 PR (#2699).** Anything that config doesn't yet satisfy is left optional and tracked as a `TODO` in the schema instead of being enforced now:
  - `hex` / `hex_contrast` are unused and dropped from the Berlin/Portland configs, but kept optional in the schema until Australia drops them too.
  - `writing_day`, `flagscheduleincomplete`, `flaghasfood`, `flaghasvisiting`, `flaghasattendeeguide`, `flaghasteam` stay optional for now.

Berlin, Portland, and the in-progress Australia 2026 configs all validate against this schema.

---
*Generated by AI*


plaindocs (commented):
Sounds good. Do we know where this is coming from?

Image

plaindocs (approved):

PR #2768, opened 2026-09-10
resize photo to standard 250 x 250

dustinloflandsmith:
note to self: learn to measure twice and cut once 🙃 

----
📚 Documentation preview 📚: https://writethedocs-www--2768.org.readthedocs.build/


plaindocs (approved):
:-D

PR #2773, opened 2026-09-15
conf: add Portland 2027 landing site with sponsorship prospectus

ericholscher:
Sponsors are planning 2027 budgets now, and the sponsorship page has been pointing them at an email address and last year's Berlin prospectus. This puts the Portland 2027 dates and prospectus on the site while the rest of the conference site stays hidden behind the landing flag until it's refreshed.

The commits are split so the first is a verbatim copy of the Portland 2026 site and the real changes sit on top of it. Review from the second commit onward. Header photos are not duplicated; 2027 pages point at the 2026 set for now. The unused cheat sheet pages are not carried over.

**Sponsorship changes worth a look**

- Keystone goes to $20,000 with a limit of 3, and the other packages scale up with it. Keystone and Patron can add activations, which are priced individually and are for booth sponsors only.
- Publisher is rebuilt as the "engage without a booth" tier: a reserved Unconference session or Writing Day project on the published schedule, job postings, and a spot on the sponsor thank-you slide. The old tier was the entry package plus two tickets and a slide logo, and sold one or two a year in Portland and zero in Berlin. Reserved sessions are for Publisher and above; the sponsor information page says so.
- Second Draft is renamed Supporter. The old name was the rung above a First Draft tier Portland dropped years ago. Config keys keep their names.
- Lightning Talks sponsorship keeps its price but requires a Supporter package or higher.

**Config instead of scrubbing**

Writing Day and Lightning Talk sponsors and the grant total used to be hardcoded in the pages. They now come from config, so the 2026 pages render exactly as before from their YAML and the 2027 copies of the same pages fall back cleanly with empty values. Event sponsors are referenced by their name in the sponsors block rather than repeated, through a small `find_sponsor` Jinja filter.

**Colour**

The palette is meant to rotate every four years, but 2025 used green out of turn, so green would have come back after two years. Coral was last used in 2023 and is the colour due, so 2027 uses the style sheet's coral. The coral icon set was missing lightning and sponsor-expo icons; they're derived from the gold ones. The coral and blue Writing Day icons had lost their even-odd fill rule and drew a solid slate sheet; that's fixed.

**Markdown**

The two remaining public rst pages on the Berlin 2026 site are converted to Markdown; the email templates stay rst.

**Things to double-check**

- The opengraph image is rendered from the built hero with the script in #2736. To get it to render here it needed the full Chromium binary with the new headless mode and the sandbox proxy passed through, and the output has a white band at the bottom from the hero's height rule at capture width. Neither change is in this PR.
- Lists in the 2027 stylesheet get the same bottom margin as paragraphs instead of non-collapsing padding. Before, every list on the prospectus ended in a gap almost twice the paragraph spacing; now list-to-text, list-to-heading, and paragraph-to-paragraph gaps all measure the same.
- CFP, grant, and Writing Day dates in the config are provisional, mirroring the 2026 timeline shifted a day earlier. They're commented as such.
- Visiting, team, and attendee guide pages are copied but their flags are off, because they still carry 2026 hotel deals, team, and logistics.

A package comparison table follows in #2776, stacked on this branch.

https://claude.ai/code/session_015JdLb6hXkD9Vym4eABf8pg


docs/_static/conf/images/headers/portland-2027-opengraph.jpg
plaindocs, line 1: These probably need a colour refresh?

plaindocs: Ah, happened a few commits down

plaindocs (approved):
I've checked the diffs sequentially, although not entirely thoroughly. I've scrolled through most of the preview.

All looks good.

docs/_static/conf/images/headers/portland-2027-opengraph.jpg
ericholscher: Yea.. in theory we will end up with a new brand for the full site, but I'm not confident of that yet :)

PR #2776, opened 2026-09-15
sponsors: add a package comparison table to the Portland 2027 prospectus

ericholscher:
## [Preview here](https://writethedocs-www--2776.org.readthedocs.build/conf/portland/2027/sponsors/prospectus/#introduction)

Stacked on #2773; only the last three commits are new here.

A sponsor reading the prospectus had to get through five prose sections before seeing a price, and the package differences that drive the decision were spread across four bullet lists. These commits reorder the page around the packages and put the differences side by side. All the wording is the existing copy, moved rather than rewritten.

**Comparison table.** Rows cascade from the benefits every package shares down to Keystone-only, so the checks form a pyramid. Add-ons sit in a labelled group underneath, ordered the same way, with their prices. Package names in the header and add-on names in the first column link to their detail. Cells spell out a value only where tiers differ in kind, such as "Premium booth" against "Booth"; everything else is a check or a dash. Prices and ticket counts come from config.

**Page order.** The opener keeps the welcome and three-day-conference sentences plus a key-facts strip, then goes straight to the table. Each package is a collapsible block showing name, price, and limit, so all four fit on one screen and open for detail. Demographics, Why Sponsor, and the no-attendee-data note now follow the packages.

**Demographics** percentages move into config and render as share bars with a legend, so the numbers have one source instead of living in prose.

**Buttons.** Two email links, one under the table and one in Inquiries, with a subject line naming the conference. The rule that centres them and gives them white text is scoped to buttons inside page content, so it also fixes the same low-contrast button on the grants, Writing Day, and volunteer pages.

**Worth checking**

- The table duplicates the bullet lists by design, so a benefit change has to land in both.
- Package detail is collapsed by default. That trades a scannable page for an extra click per package, and linking to a package scrolls to a closed block.
- The Benefits headings inside the collapsible blocks are plain HTML, because a Markdown heading there opens a docutils section that escapes the surrounding element.
- Below tablet width the table scrolls sideways with the benefit column pinned. Worth a look on a real device.

https://claude.ai/code/session_015JdLb6hXkD9Vym4eABf8pg


plaindocs (approved):
This is great.

Logistically, 30s vs 60s intros is gonna be hard to manage. otherwise :shipit:

ericholscher:
> This is great.
> 
> Logistically, 30s vs 60s intros is gonna be hard to manage. otherwise :shipit:

Yea.. in practice we won't enforce it, but it looks better in the prospectus :D

PR #2777, opened 2026-09-15
docs: speed up the Sphinx build

ericholscher:
A full site build took about 147 seconds locally (4 cores) and is now about 43 seconds, with 63 seconds when limited to two workers, which is closer to a Read the Docs builder. Output is byte-identical to the previous build except for feed timestamps, the header image the 2025 template picks at random, and which of two same-named images gets the `_images/foo.jpg` vs `foo1.jpg` name, which now depends on parallel read order.

Profiling found three things dominating, none of them "we have a lot of content":

- The sidebar's `toctree(collapse=False)` made Sphinx resolve the entire site toctree (thousands of nodes) for every page that renders the sidebar, then prune it to eight items. That was most of the write phase and nearly all of ablog's archive-page phase.
- MyST `eval-rst` blocks parse in a temporary document whose settings still point at the build environment, so every page using them pickled the 20MB environment into its doctree. Doctrees totalled 500MB and lived inside the uploaded output directory.
- ablog declares itself unsafe for parallel reading, which forced a serial read of 1,408 documents. Making it parallel-safe needed more than the merge handler: ablog's stored post copies kept each worker's environment alive, so transfers snowballed to gigabytes until those copies were detached and workers stopped sending back inherited posts.

All of that lives in a new `docs/_ext/build_perf.py`, with each fix independent and explained in place. It patches two private Sphinx toctree functions, which is fine within the pinned `Sphinx>=8.1,<9` but should be rechecked on a Sphinx 9 upgrade.

Two smaller things ride along: PyStemmer (a C stemmer snowballstemmer picks up automatically) makes search indexing cheaper with an identical index, and five dependencies only used by `docs/_scripts/` move to a non-default `scripts` uv group so Read the Docs installs 29 fewer packages.

Worth a look from a reviewer: the parallel read changes which document "wins" for duplicate ablog labels such as `announcing-cfp`, since that now follows merge order instead of alphabetical read order. None of the site's `:ref:` targets are among the duplicates.

https://claude.ai/code/session_01PR7bnxqKzVS6GAaSKaBMc2

---
_Generated by [Claude Code](https://claude.ai/code/session_01PR7bnxqKzVS6GAaSKaBMc2)_

----
📚 Documentation preview 📚: https://writethedocs-www--2777.org.readthedocs.build/


ericholscher:
@mxsasha @plaindocs not sure if either of y'all can really review this (I can barely do a high-level and my Sphinx knowledge is likely much deeper :D) -- but seems worth trying to see if we can magically get faster builds.

plaindocs:
Very much in favor of this. As you say, can't easily read the sphinx python stuff.

But if we're diffing the output and it looks fine it should be safe enough?

Do we also need to do anything to the GHA builds?

> Worth a look from a reviewer: the parallel read changes which document "wins" for duplicate ablog labels such as announcing-cfp, since that now follows merge order instead of alphabetical read order. None of the site's :ref: targets are among the duplicates.

if we cared about this we could add years to labels to we dedupe automatically?

ericholscher:
Let's see if anything blows up :)

PR #2778, opened 2026-09-16
conf: freeze conferences up to 2024 as static HTML

ericholscher:
Stacked on #2777; review that one first, and this diff is only readable with the base set to its branch.

Conferences through 2024 are over and their pages never change, but Sphinx was still parsing and rendering all 825 of them on every build, 59% of the site. They now ship as pre-rendered HTML under `_static_html`, served by `html_extra_path`, the same arrangement 2015-2017 have had for years. On top of #2777 the build drops from 36 to 18 seconds locally.

Every conference page keeps its existing URL and the 29 pre-existing broken internal links on the site are unchanged, neither more nor fewer, which I checked by diffing the full set of resolved internal references before and after. htmlproofer passes.

What actually goes away is the blog tag archives for those conferences, such as `/blog/archive/tag/portland-2019/`, because the news pages stop being ablog posts. That was the one real cost and you said it doesn't matter. Nothing else disappears: 40 pages vanish from the output and all 40 are those tag archives.

The part worth a careful look is images. Sphinx numbers generated image filenames per build, so `volunteer.jpg` and `volunteer1.jpg` can swap between builds, and a frozen page pointing at `_images/` would silently show the wrong picture later. Frozen pages now load from `_static_html/_frozen-images/` instead, which is immune to that. Finding this also turned up three images on the existing 2015-2017 pages that only rendered because some 2018-2024 page happened to reference the same file; those would have broken here, and are fixed.

Two things a reviewer should weigh:

- The repository grows by about 32MB, since rendered HTML is far bulkier than source. Read the Docs clones that on every build, which claws back a little of the win.
- Frozen pages are now snapshots. They keep today's sidebar, footer and CSS and will drift as the live site changes. That is inherent to freezing and already true of 2015-2017.

Per-year YAML in `docs/_data/` is deliberately left in place. The build no longer reads it, but it is the historical record for those conferences and keeping it makes the freeze reversible. `AGENTS.md` documents the convention and how to freeze the next year.

https://claude.ai/code/session_01PR7bnxqKzVS6GAaSKaBMc2

---
_Generated by [Claude Code](https://claude.ai/code/session_01PR7bnxqKzVS6GAaSKaBMc2)_

----
📚 Documentation preview 📚: https://writethedocs-www--2778.org.readthedocs.build/


plaindocs:
What about archiving the conference but leaving the blog posts?

ericholscher:
I don't think we care about the blog post archives?

